PR #46, opened 2020-09-16
feat: export parser interfaces in Interfaces

RasPhilCo:


src/interfaces/parser.ts
```diff
@@ -0,0 +1,171 @@
+import {AlphabetLowercase, AlphabetUppercase} from './alphabet'
+import {Config} from './config'
+
+export type ParseFn<T> = (input: string) => T
+
+// eslint-disable-next-line @typescript-eslint/interface-name-prefix
```
amphro, line 6: This eslint diable is no longer needed right?

src/interfaces/parser.ts
```diff
@@ -0,0 +1,171 @@
...
+  required: false;
+  value?: T;
+}
+
+export type ParserArg<T> = RequiredArg<T> | OptionalArg<T>
+
+export interface FlagOutput { [name: string]: any }
+export type ArgInput = Arg<any>[]
+
+// eslint-disable-next-line @typescript-eslint/interface-name-prefix
```
amphro, line 42: This is no longer needed?

amphro (approved):

src/interfaces/parser.ts
RasPhilCo, line 6:
```suggestion
```

src/interfaces/parser.ts
RasPhilCo:
```suggestion
```

PR #101, opened 2021-01-07
feat: add global flags

RasPhilCo:
WIP

requires #94 #99

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
   "oclif": {
     "bin": "oclif",
     "plugins": [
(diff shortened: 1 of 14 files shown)


RodEsp:
I believe this functionality has already been added with the addition of the `--json` global flag: https://github.com/oclif/core/blob/11cbd6ba8caad6a8df814fec55be45b05f52dd4b/src/command.ts#L115-L120.

Do we still need this PR for anything or should we close it @RasPhilCo?

PR #111, opened 2021-02-23
feat: add topic separator option

RasPhilCo:


src/config/config.ts
```diff
@@ -298,7 +301,8 @@ export class Config implements IConfig {
   }
 
   get commandIDs() {
-    return uniq(this.commands.map(c => c.id))
+    const ids = Lodash.flattenDeep(this.commands.map(c => [c.id, c.aliases]))
```
RasPhilCo, line 304: @amphro let me know your thoughts, but here I'm saying `commandIDs` should return alias ids as well

RasPhilCo: maybe some downstream affects I'm not thinking through

amphro (approved):
Nice! A few more tests around finding commands might be nice.

src/help/util.ts
```diff
@@ -39,3 +39,32 @@ export function template(context: any): (t: string) => string {
...
+
+function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
+  if (argv.length === 1) return argv
+
+  const ids = config.commandIDs.concat(config.topics.map(t => t.name))
+
+  const findId = (id: string, next: string[]): string | undefined => {
+    const idPresnet = (id: string) => ids.includes(id)
+    if (idPresnet(id) && !idPresnet(`${id}:${next[0]}`)) return id
+    if (next.length === 0 || next[0] === '--') return
```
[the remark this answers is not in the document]
RasPhilCo: This is returning no match b/c the line would return the if was the last valid id found. This line is saying, you didn't find anything to return (line above would of returned) and you have no further argv to explore.

PR #129, opened 2021-03-16
fix: don't resolve lib -> src in development mode

RasPhilCo:
~~WIP do not merge~~

This change thus introduces the idea of explicit dev mode in oclif.

This also means we are now introducing two separate bin stubs for dev & prod `bin/dev` and `bin/run`.

bin/dev:
```
#!/usr/bin/env node

process.env.NODE_ENV = 'development'
const path = require('path')
const project = path.join(__dirname, '../tsconfig.json')
require('ts-node').register({project})
require('@oclif/core').run()
.then(require('@oclif/core/flush'))
.catch(require('@oclif/core/handle'))
```

bin/run:
```
#!/usr/bin/env node

require('@oclif/core').run()
.then(require('@oclif/core/flush'))
.catch(require('@oclif/core/handle'))

```

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -26,15 +26,6 @@ function channelFromVersion(version: string) {
   return (m && m[1]) || 'stable'
 }
 
-function hasManifest(p: string): boolean {
-  try {
-    require(p)
-    return true
-  } catch {
-    return false
-  }
-}
-
 const WSL = require('is-wsl')
 
 function isConfig(o: any): o is IConfig {
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -181,9 +181,8 @@ export class Config implements IConfig {
   }
 
   async loadDevPlugins() {
-    if (this.options.devPlugins !== false) {
-      // do not load oclif.devPlugins in production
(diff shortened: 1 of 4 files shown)


src/config/config.ts
```diff
@@ -181,9 +172,8 @@ export class Config implements IConfig {
   }
 
   async loadDevPlugins() {
-    if (this.options.devPlugins !== false) {
-      // do not load oclif.devPlugins in production
-      if (hasManifest(path.join(this.root, 'oclif.manifest.json'))) return
+    // do not load oclif.devPlugins in production
+    if (this.options.devPlugins !== false && process.env.NODE_ENV === 'development') {
```
amphro, line 176: Why would devPlugins be false?

Can you put the env check in an `isProd()` method?

amphro (approved):

joelday:
Side question: Does this impact ts-node running for linked plugins?

RasPhilCo:
Pardon @joelday, missed your comment earlier, @amphro is releasing this today.

amphro:
I never commented but I did the release.

PR #143, opened 2021-04-09
feat: support misspelled spaced commands

mdonnalley:
Add support for misspelled spaced commands

### DIFF
--- a/src/help/util.ts
+++ b/src/help/util.ts
@@ -45,14 +45,29 @@ function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
 
   const ids = config.commandIDs.concat(config.topics.map(t => t.name))
 
-  const findId = (id: string, next: string[]): string | undefined => {
+  const findId = (argv: string[]): string | undefined => {
+    const final: string[] = []
     const idPresent = (id: string) => ids.includes(id)
-    if (idPresent(id) && !idPresent(`${id}:${next[0]}`)) return id
-    if (next.length === 0 || next[0] === '--') return
-    return findId(`${id}:${next[0]}`, next.slice(1))
+    const isFlag = (s: string) => s.startsWith('-')
+    const finalizeId = (s?: string) => s ? [...final, s].join(':') : final.join(':')
+
+    const hasSubCommandsWithArgs = () => {
+      const subCommands = config.commands.filter(c => (c.id).startsWith(finalizeId()))
+      return Boolean(subCommands.find(cmd => cmd.args.length > 0))
+    }
+
+    for (const arg of argv) {
+      if (idPresent(finalizeId(arg))) final.push(arg)
+      // If the parent topic has a command that expects positional arguments, then we cannot
+      // assume that any subsequent string could be part of the command name
+      else if (isFlag(arg) || hasSubCommandsWithArgs()) break
+      else final.push(arg)
+    }
+
+    return finalizeId()
   }
 
-  const id = findId(argv[0], argv.slice(1))
+  const id = findId(argv)
 
   if (id) {
     const argvSlice = argv.slice(id.split(':').length)
--- a/src/help/util.ts
+++ b/src/help/util.ts
@@ -46,8 +46,8 @@ function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
   const ids = config.commandIDs.concat(config.topics.map(t => t.name))
 
   const findId = (id: string, next: string[]): string | undefined => {
-    const idPresnet = (id: string) => ids.includes(id)
-    if (idPresnet(id) && !idPresnet(`${id}:${next[0]}`)) return id
+    const idPresent = (id: string) => ids.includes(id)
+    if (idPresent(id) && !idPresent(`${id}:${next[0]}`)) return id
     if (next.length === 0 || next[0] === '--') return
     return findId(`${id}:${next[0]}`, next.slice(1))
   }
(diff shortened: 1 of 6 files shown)


src/help/util.ts
```diff
@@ -46,8 +46,8 @@ function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
   const ids = config.commandIDs.concat(config.topics.map(t => t.name))
 
   const findId = (id: string, next: string[]): string | undefined => {
-    const idPresnet = (id: string) => ids.includes(id)
-    if (idPresnet(id) && !idPresnet(`${id}:${next[0]}`)) return id
+    const idPresent = (id: string) => ids.includes(id)
+    if (idPresent(id) && !idPresent(`${id}:${next[0]}`)) return id
```
RodEsp, line 50: Lulz

src/main.ts
```diff
@@ -55,7 +56,24 @@ export async function run(argv = process.argv.slice(2), options?: Interfaces.Loa
   const cmd = config.findCommand(id)
   if (!cmd) {
     const topic = config.findTopic(id)
-    if (topic) return config.runCommand('help', [id])
+    if (topic) {
+      const subCommands = config.commands.filter(c => (c.id).startsWith(id))
+      const argsOfSubCommands = subCommands.reduce((x, y) => x.concat(y.args), [] as Arg[])
```
RodEsp, line 61: What's the purpose of this reduce? Is it not possible to just cast subCommands to Arg[]?

Not sure of the right TS syntax but something like `const argsOfSubCommands = subCommands as Array<Arg>` or `const argsOfSubCommands = <Array<Arg>>subCommands`?

src/main.ts
```diff
@@ -55,7 +56,24 @@ export async function run(argv = process.argv.slice(2), options?: Interfaces.Loa
...
+
+      if (argsOfSubCommands.length > 0) return config.runCommand('help', [id])
+      if (argvSlice[0].startsWith('--')) return config.runCommand('help', [id])
```
RodEsp, line 64:
```suggestion
      if (argsOfSubCommands.length > 0 || argvSlice[0].startsWith('--')) return config.runCommand('help', [id])
```

src/main.ts
mdonnalley: @RodEsp The point is to get a flat list of all args across all the sub commands.

`subCommands` essentially looks like this:

```
[
  { args: [arg1, arg2] },
  {args: [arg3, arg4] }
]
```

And I want this: `[arg1, arg2, arg3, arg4]` which is why I used `reduce`


Regarding the `as Arg[]`: that was required to get the compiler to be okay with accessing the `arg` property within the reducer

src/main.ts
```diff
@@ -55,7 +56,23 @@ export async function run(argv = process.argv.slice(2), options?: Interfaces.Loa
   const cmd = config.findCommand(id)
   if (!cmd) {
     const topic = config.findTopic(id)
-    if (topic) return config.runCommand('help', [id])
+    if (topic) {
+      const subCommands = config.commands.filter(c => (c.id).startsWith(id))
+      const argsOfSubCommands = subCommands.reduce((x, y) => x.concat(y.args), [] as Arg[])
+
+      if (argsOfSubCommands.length > 0 || argvSlice[0].startsWith('--')) return config.runCommand('help', [id])
```
amphro, line 63: Why does the first arg starting with `--` justify help?

src/main.ts
amphro: Ya, that `as Arg[]` is very standard. 

The reduce makes sense to flatten, but you are just checking the length? Is the question, "is there one matching command with arguments?" Can't you just say `!!subCommands.find(cmd => cmd.args.length > 0)`

src/main.ts
```diff
@@ -55,7 +56,23 @@ export async function run(argv = process.argv.slice(2), options?: Interfaces.Loa
...
+
+      const extraBits = []
+      for (const arg of argvSlice) {
+        if (arg.startsWith('--')) {
```
amphro, line 67: Do short chars using `-` matter?

src/main.ts
```diff
@@ -55,7 +56,23 @@ export async function run(argv = process.argv.slice(2), options?: Interfaces.Loa
...
+          break
+        } else {
+          extraBits.push(arg)
+        }
+      }
+      id = [id, ...extraBits].join(':')
```
amphro, line 73: So basically you are trying to find arguments before the first defined flag and create a new command id. Shouldn't this be what `findCommand` does?

This is hard for me to follow. Can you add some more info in some comments, tests, or the PR description? Maybe we should meet to discuss.

amphro (changes requested):

src/help/util.ts
```diff
@@ -45,14 +45,29 @@ function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
 
   const ids = config.commandIDs.concat(config.topics.map(t => t.name))
 
-  const findId = (id: string, next: string[]): string | undefined => {
-    const idPresnet = (id: string) => ids.includes(id)
-    if (idPresnet(id) && !idPresnet(`${id}:${next[0]}`)) return id
-    if (next.length === 0 || next[0] === '--') return
-    return findId(`${id}:${next[0]}`, next.slice(1))
+  const findId = (argv: string[]): string | undefined => {
```
amphro, line 48: Nice, I think this is really clean.

amphro (approved):

Issue #130, opened 2021-03-25
ES Module support for Oclif v2


typhonrt:
Hi Philipe & Thomas (@RasPhilCo / @amphro) et al,

Great work on Oclif v2 Philipe and all involved. I really like the refactor to a core module compared to v1. Removing the circular dependencies (IE between v1 plugin-help and plugin-command) helps a bunch. I would like to discuss adding ES Module (ESM) support to Oclif v2 before initial launch in ~June. I initially prototyped the changes in v1 and saw the core announcement and have fully worked out the essential changes to @oclif/core for v2. I have a fork of @oclif/core with these changes and have fully implemented non-trivial Oclif middleware and resulting CLI w/ ESM to test everything out that is published on NPM using my @oclif/core fork. I also have three test repositories that have stripped down CLI tests with variations that respectively cover a wide swath of the 12.0.0 to 15.x Node ecosystem by Github Actions. I'll comment on this below in another post.

This kind of fundamental change is best in a new version of Oclif and is timely due to the minimum Node support of v12 for Oclif v2 which is the first Node LTS version that has wider support for ES Modules. Supporting ESM now future proofs Oclif as the larger Node ecosystem transitions to ESM in the coming years. There is a fair amount of movement in this direction already in April / May as Node 12 becomes the minimum LTS release. These changes do not affect current support of CommonJS / Typescript CLIs. In fact there could be some benefit for end user CLI development in Typescript as one could target ESM instead of CJS and if launching a CLI w/ Node 14 as a minimum version for support can target ES2020 for Typescript or simply write the CLI in ESM / ES2020. The main core Oclif v2 code should still of course target CJS for now, but this frees up end users to choose their target of choice between CJS / ESM / Typescript. 

I will go over in more detail in a follow up post to this issue regarding what the changes cover and note the other areas of the larger codebase / other modules that require changes. I have already gone through several iterations of improving the essential changes and have arrived at what I think is a clean addition which I have tested across all the OS platforms w/ Node `12.0.0` - `15.x`. Beyond just submitting the pull requests / changes I am fully committed to helping write new tests / verifying Node 12.0.0 compatibility and assisting with writing documentation changes for v2. IE not just provide a code dump and add extra burden to any internal team members before the v2 launch. I believe there is enough time to get these changes in responsibly before the v2 ~June launch. 

[...]

PR #144, opened 2021-04-14
Added ESM loading support - tests pass - MacOS / Linux / Windows works

amphro:
See https://github.com/oclif/core/issues/130 for more information. Creating the PR for comments, testing, and further discussion.


src/config/plugin.ts
```diff
@@ -73,6 +74,8 @@ export class Plugin implements IPlugin {
 
   pjson!: PJSON.Plugin
 
+  isESM = false
```
amphro, line 77: Can't we completely hide this in ModuleLoader and just call `ModuleLoader.load()` here instead? The `getPackageType` call in the ModuleLoader code should resolve the correct pjson.type. To potentially reduce file io, you could also pass in pjson.type to the `ModuleLoader.load` method maybe?

amphro:
@typhonrt I did an initial review. Let me know your thoughts.

typhonrt:
Apologies the delay. I was deep in refactoring several core modules for ESM / ES2020 / 21 support. I'm readily available going forward through the Oclif v2 launch. I'll answer all queries from above inline:

>We don't have a @oclif/core version of the docs yet, but we will have to update the reference to hook execution order when we do. https://github.com/oclif/oclif.github.io/blob/docs/docs/hooks.md

The hook execution order is the same with the changes (everything happens in sequential order). Documentation additions should cover that you can resolve an actual package name as a hook or from Node 12.16+ which supports the exports property in `package.json`  a particular export w/ a defined path after the package name. 

> Can we change this to fetchHelpClass now that it is async?

I'm definitely open to all changes regarding name, structure, and formatting; glad to get this code into Oclif v2!

> I don't love shipping with 0.x releases. I took a look in the repo and I can't tell if there are plans to have a 1.x release. They might be waiting for a node equivalent?

Re: [get-package-type](https://www.npmjs.com/package/get-package-type), indeed is a bit of stopgap module examining the source though it is being used by `nyc` and has many downstream dependents (nearly 500k) and is doing 7MM downloads a month. It does work and is a CJS module and can be easily used by Oclif v2 being a TS project targeting CommonJS. 

I didn't want to be so brazen as to try and include my own module that handles the same task. I have a significantly more robust module that is built for easier ESM usage, IE supports file URLs / `import.meta.url`. It certainly doesn't have the downstream consumers yet and `0.1.0` for current package version too; I'll be bumping all my new ESM modules to `1.0.0` in the coming month. There is thorough documentation and lots of recent activity shoring it up: [@typhonjs-utils/package-json](https://github.com/typhonjs-node-utils/package-json)

Even if a Node API equivalent became available this wouldn't be available Node 12+, so some solution like `get-package-type` or my module is necessary. 

> I think this name is a little weird, maybe loadWithData or loadWithInfo?

I definitely like the `WithData` suggestion and adopted a similar pattern in my own module mentioned above. 

> Can't we completely hide this in ModuleLoader and just call ModuleLoader.load() here instead? The getPackageType call in the ModuleLoader code should resolve the correct pjson.type. To potentially reduce file io, you could also pass in pjson.type to the ModuleLoader.load method maybe?

I do agree that this can be moved to ModuleLoader. In this case it might be best to check the `pjson.type` directly from the config instead of calling getPackageType for every hook. This can be cleaned up and `isESM` in the config.ts / plugin.ts removed. 

--------

Now I think the best way to proceed is to actually make a new PR. I just want to be absolutely sure there is a clean merge process since there are some conflicting files now with the main branch. I do not see any particular overlapping conflict. 

I'll cleanup all of the suggested changes above in addition to adding the quality of life changes I mentioned in #130. 

The one area I would like to get your input is in the help subsystem going fully asynchronous. This change is not in the initial PR as also mentioned in #130.  While I rewrote [./src/help/index.ts](https://github.com/oclif/core/blob/master/src/help/index.ts) in ESM I have fully prototyped these changes in my middleware as you can see [here](https://github.com/typhonjs-oclif/core/blob/main/src/system/help/Help.js#L123). The changes are for `showHelp`, `showCommandHelp`, `showRootHelp`, and `showTopicHelp` to be made async. I'd be glad to discuss these changes in more detail as necessary, but would like to include them in the new PR. 

I'll get started on the new feature branch and test things tomorrow. Do let me know if this sounds like a good way to proceed. 

Regards...

amphro:
> I'll cleanup all of the suggested changes above in addition to adding the quality of life changes I mentioned in #130.

Sounds good. I'll close this PR. If I remember correctly, I agree with the quality of life changes but it might be better to do separate PRs: one for the ESM then one for quality of life. It will make the PRs easier to digest and focused if they don't depend on each other. 

> The one area I would like to get your input is in the help subsystem going fully asynchronous.

I don't see a problem with this. Let's definitely keep this in a separate PR. We will also be adding / changing several help class features in the next month, so I can also make these changes if you want.

gotjoshua:
many thanks for this enhancement! 
I was so confused when i couldn't import stuff in version 1.x

PR #160, opened 2021-05-07
Integrate ESM loading of commands & hooks

typhonrt:
Greets. This is the first of three PRs regarding adding ESM support to Oclif v2. 

This initial PR adds ESM loading of commands and hooks. I have tested it with my ESM Oclif middleware / CLI tool in addition to all tests passing in the repo. 

The easiest way for me to proceed with working on the Help subsystem adding ESM help class loading support and async handling is for this PR to be merged first as it depends on the ModuleLoader addition. 

Let me know your thoughts @amphro.


src/config/plugin.ts
```diff
@@ -169,23 +170,24 @@ export class Plugin implements IPlugin {
...
         return Object.values(cmd).find((cmd: any) => typeof cmd.run === 'function')
       }
+      const isESM = this.pjson.type === 'module'
       const p = require.resolve(path.join(this.commandsDir, ...id.split(':')))
-      this._debug('require', p)
+      this._debug(isESM ? '(import)' : '(require)', p)
       let m
       try {
-        m = require(p)
+        m = isESM ? await ModuleLoader.importDynamic(p) : require(p)
```
amphro, line 190: Can you call `ModuleLoader.load(this, p)` since that has the importDynamic/require logic?

src/module-loader.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const getPackageType = require('get-package-type')
+
+/**
+ * Provides a mechanism to use dynamic import / import() with tsconfig -> module: commonJS as otherwise import() gets
+ * transpiled to require().
+ */
+const _importDynamic = new Function('modulePath', 'return import(modulePath)') // eslint-disable-line no-new-func
+
+export default class ModuleLoader {
```
amphro, line 16: I thought your original PR had tests? Is this worth unit testing?

amphro:
I've tried this branch on a couple of oclif/core cli's that we have and things look good. I haven't tested on a distributed cli with plugins installed but @oclif/plugin-plugins hasn't been updated yet to use core anyways. I'm good to merge this once I get your responses to my comments.

src/module-loader.ts
typhonrt: There were no added tests. I'm certainly willing to submit tests as well.

src/config/plugin.ts
typhonrt: Both load and loadWithData pass the string through ModuleLoader.resolvePaths

This require.resolve line  (`require.resolve(path.join(this.commandsDir, ...id.split(':')))`) uses `this.commandDir`
https://github.com/oclif/core/pull/160/files/28566d746b4ec5b0d8be3ef2ba5d7808d213452d#diff-0edcef2f1da996cde64784a30d76043fbe58fb9352f6b0189e1d135b868ee8f1L184

In ModuleLoader.resolvePaths `config.root` is used:
`filePath = isESM ? require.resolve(path.join(config.root, modulePath))`

---------

However this could be a solution (tests pass with this below):
```js
      let m
      try {
        const p = path.join(this.pjson.oclif.commands as string, ...id.split(':'))
        const {isESM, module, filePath} = await ModuleLoader.loadWithData(this, p)
        this._debug(isESM ? '(import)' : '(require)', filePath)
        m = module
      } catch (error) {
        if (!opts.must && error.code === 'MODULE_NOT_FOUND') return
        throw error
      }

```

typhonrt: A further note is that since `error.code` is being checked in the catch statement that it should be verified that the code is the same for require and dynamic import. My assumption is that may not be the case.

typhonrt: @amphro It turns out in a CommonJS environment that both require and import dynamic return an error code of `MODULE_NOT_FOUND` so that part is fine.  

It is interesting to note that in an ESM environment import dynamic has an error code of `ERR_MODULE_NOT_FOUND` and require through module.createRequire returns `MODULE_NOT_FOUND`.

typhonrt: I have to say I'm certainly shying away from any custom error wrapping of the loading process. `MODULE_NOT_FOUND`/ `ERR_MODULE_NOT_FOUND` is one of many errors that can be thrown; IE various errors in the code being loading itself should be posted.

So, I have updated the PR with what I think is best. 

For reference I have implemented my own dual Node / browser module loading code as an ESM Node module that is similar to what is being discussed here (has test cases / test framework):
https://github.com/typhonjs-node-utils/loader-module

amphro (approved):
I like the changes! I liked the idea of ModuleLoader throwing a custom error for module not found, but I don't really care. Catching both is totally fine. In the case of a custom error, I definitely wouldn't try to wrap every type of error either. You could also put a comment why the two types of errors.

src/module-loader.ts
amphro: At the bare minimum, can you add a new [fixture for a esm](https://github.com/oclif/core/tree/master/test/command/fixtures) and a [main test for it here](https://github.com/oclif/core/blob/master/test/command/main.test.ts#L76).

amphro:
After you get the tests in, I'll merge this in.

typhonrt:
> I like the changes! I liked the idea of ModuleLoader throwing a custom error for module not found, but I don't really care. Catching both is totally fine. In the case of a custom error, I definitely wouldn't try to wrap every type of error either. You could also put a comment why the two types of errors.

I suppose that is true regarding only handling the two module not found error codes then passing through everything else. I'm thinking a custom CLIError is the way to go. I'll put it in `./src/errors/errors` and export it through `./src/errors/index.ts`

This is what I'm thinking:
```js
import {CLIError} from './cli'
import {OclifError} from '../../interfaces'

export class ModuleLoadError extends CLIError implements OclifError {
  oclif!: { exit: number }

  code = 'MODULE_NOT_FOUND'

  constructor(message: string) {
    super(`[MODULE_NOT_FOUND] ${message}`, {exit: 1})
    this.name = 'ModuleLoadError'
  }

  render(): string {
    return ''
  }
}
```

I do think having a custom error makes a lot of sense for Oclif because the runtime is only Node / V8 and it provides a concise and unified error stack leading back to ModuleLoader with one defined error for require or dynamic import when the module is not found. The errors printed out are quite a bit different and verbose in the case of require versus dynamic import and it easily could confuse someone not accustomed to knowing what to expect in advance. This is what the error message will be like (paths substituted):

```
     ModuleLoadError: [MODULE_NOT_FOUND] require failed to load /some/path/to/cli/bad-path.js
      at Function.loadWithData (node_modules/@oclif/core/lib/module-loader.js:64:15)
      at fetch (node_modules/@oclif/core/lib/config/plugin.js:188:62)
      at Plugin.findCommand (node_modules/@oclif/core/lib/config/plugin.js:201:23)
      at Object.load (node_modules/@oclif/core/lib/config/plugin.js:131:54)
      at Config.runCommand (node_modules/@oclif/core/lib/config/config.js:261:29)
      at Object.run (node_modules/@oclif/core/lib/main.js:64:16)
```

for dynamic import: `ModuleLoadError: [MODULE_NOT_FOUND] import() failed to load /some/path/to/cli/bad-path.js`

amphro:
Ya, I think that looks good. How come you empty out the render?

> dynamic import error has an internal Node stack that gives no indication of where the error actually occurred

Why is that? Wouldn't creating the new CLIError create a new stack trace from the module loader class?

typhonrt:
@amphro This PR should be good to merge now. I added ModuleLoadError to handle the CJS & ESM loader variance with the module not found case and added reasonably thorough unit tests for ModuleLoader in `./test/module-loader`.

amphro:
Love the changes. Nice work on this.

PR #166, opened 2021-05-27
ESM quality of life improvements

typhonrt:
@amphro 
This rounds out ESM support for @oclif/core v2. 

ESM quality of life improvements:
- support file URL string as options input (IE 'import.meta.url').
- added `./flush.js` as an export from @oclif/core; `./src/index.ts`
- added main runtime ESM test w/ file URL as options.
- raised package.json Node engine level to `>= 12.0.0`. 

The added runtime tests focus on Main.run using a file URL as options. I also changed one of the Config ESM tests to use file URL as options as well. 

Please see the follow up message for a note about CommonJS deprecations when I raised the engine level to `>= 12.0.0`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,6 +13,8 @@ import * as Parser from './parser'
 import {Hook} from './interfaces/hooks'
 import {settings, Settings} from './settings'
 
+const flush = require('../flush')
+
 export {
   Command,
   Config,
@@ -32,6 +34,7 @@ export {
   toConfiguredId,
   settings,
   Settings,
+  flush,
 }
 
 function checkCWD() {
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "typescript": "3.8.3"
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=12.0.0"
   },
   "files": [
     "/lib",
--- a/src/command.ts
+++ b/src/command.ts
@@ -1,3 +1,5 @@
+import {fileURLToPath} from 'url'
+
 import {format, inspect} from 'util'
 
 import {Config} from './config'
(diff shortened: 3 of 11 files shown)


typhonrt:
Regarding the CommonJS deprecations please see the following image which popped up in my IDE / WebStorm:

Screen Shot 2021-05-26 at 10 11 37 PM

It appears after Node 12.19.0 and 14.6.0 that `module.parent` is deprecated. The suggestion is to use `require.main` or `module.children` instead. Since this is a CommonJS change to Oclif I did not make any modifications. It might be prudent to track this in a separate issue and work out any changes necessary so that deprecated functionality isn't depended on for future Node versions. 

Relevant Node documentation:
https://nodejs.org/api/modules.html#modules_module_parent

src/index.ts
```diff
@@ -13,6 +13,8 @@ import * as Parser from './parser'
 import {Hook} from './interfaces/hooks'
 import {settings, Settings} from './settings'
 
+const flush = require('../flush')
```
amphro, line 16: While you are here @typhonrt - can you also do the same thing with `../handle`?

amphro (approved):

src/index.ts
typhonrt: I thought about that, but handle is accessible through Errors which is already exported with handle. (below uses the CommonJS workaround for named exports):

```js
#!/usr/bin/env node
import oclif from '@oclif/core';

oclif.run(void 0, import.meta.url)
.then(oclif.flush)
.catch(oclif.Errors.handle);
```

My general thought is that `./handle.js` should be removed as it is redundant. I should have brought this up.

amphro: Fair enough.

PR #180, opened 2021-06-03
Add global --json flag

mdonnalley:


src/command.ts
```diff
@@ -186,4 +222,12 @@ export default abstract class Command {
       console.error(error)
     }
   }
+
+  protected toSuccessJson(result: Record<string, any>): Record<string, any> {
+    return {result}
```
amphro, line 227: Just return result? People transitioning from `sfdx` -> `sf` would need to remove the `.result` but I think that is cleaner.

src/command.ts
```diff
@@ -108,18 +126,23 @@ export default abstract class Command {
...
   async _run<T>(): Promise<T | undefined> {
     let err: Error | undefined
+    let result
     try {
       // remove redirected env var to allow subsessions to run autoupdated client
       delete process.env[this.config.scopedEnvVarKey('REDIRECTED')]
-
       await this.init()
-      return await this.run()
+      result = await this.run()
```
amphro, line 134: The method `toSuccessJson` expects `Record<string, any>`. If we want to keep that, then you need to update line 179 to `abstract run(): PromiseLike<T>` where T is defined on command: `export default abstract class Command<T extends Record<string, any> = unknown>{`

Otherwise, you could update toSuccessJson to accept any like the run method returns now.

amphro (changes requested):
In the PR description, you say that "Plugins" can disable it, but that isn't exactly true. I think we should say "CLIs" because if it is set globally, it will turn it off for all of oclif for that process. So plugins installed in LCIs shouldn't touch the settings object.

amphro (approved):

RodEsp (approved):

PR #181, opened 2021-06-04
Add `url` flag

mdonnalley:
This PR adds two new flag types:
- `url`

@W-9320561@

### DIFF
--- a/src/parser/flags.ts
+++ b/src/parser/flags.ts
@@ -60,12 +60,13 @@ export const url = build({
  */
 export const path = build({
   parse: async input => {
+    if (input === '<cwd>') return process.cwd()
     const invalidPathChars = /[["?<>|\]]+/
     if (invalidPathChars.test(input))
       throw new Error(`Expected a valid file path but received: ${input}`)
     return input
   },
-  default: process.cwd(),
+  default: '<cwd>',
 })
 
 export function option<T>(
--- a/src/parser/flags.ts
+++ b/src/parser/flags.ts
@@ -45,18 +45,19 @@ export const url = build({
     try {
       return new URL(input)
     } catch {
-      throw new Error(`Expected a valid url but received ${input}`)
+      throw new Error(`Expected a valid url but received: ${input}`)
     }
   },
 })
 
-export const filepath = build({
+export const path = build({
   parse: async input => {
     const invalidPathChars = /[["?<>|\]]+/
     if (invalidPathChars.test(input))
-      throw new Error(`Expected a valid file path but received ${input}`)
+      throw new Error(`Expected a valid file path but received: ${input}`)
     return input
   },
+  default: process.cwd(),
 })
 
 export function option<T>(
--- a/src/parser/flags.ts
+++ b/src/parser/flags.ts
@@ -40,6 +40,10 @@ export const integer = build({
   },
 })
 
+/**
+ * Initializes a string as a URL. Throws an error
+ * if the string is not a valid URL.
+ */
 export const url = build({
   parse: async input => {
     try {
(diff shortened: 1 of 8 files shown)


src/parser/flags.ts
```diff
@@ -40,6 +40,25 @@ export const integer = build({
   },
 })
 
+export const url = build({
+  parse: async input => {
+    try {
+      return new URL(input)
+    } catch {
+      throw new Error(`Expected a valid url but received ${input}`)
```
amphro, line 48: Add a `:` to make it consistent with the integer error?

src/parser/flags.ts
```diff
@@ -40,6 +40,25 @@ export const integer = build({
...
+    }
+  },
+})
+
+export const filepath = build({
+  parse: async input => {
+    const invalidPathChars = /[["?<>|\]]+/
+    if (invalidPathChars.test(input))
+      throw new Error(`Expected a valid file path but received ${input}`)
```
amphro, line 57: ditto

src/parser/flags.ts
```diff
@@ -40,6 +40,25 @@ export const integer = build({
...
+  parse: async input => {
+    try {
+      return new URL(input)
+    } catch {
+      throw new Error(`Expected a valid url but received ${input}`)
+    }
+  },
+})
+
+export const filepath = build({
```
amphro, line 53: What if we just called it `path`? That way it doesn't matter if it is a file or directory?

I'm also thinking what else we could do here that would be interesting to make it a little more helpful than just string. For example, wanting a path that defaults to the cwd is super useful, but the help gets all messed up. Maybe that is something this flag could fix? i.e. if `default: <cwd>` then automatically initialize the value to `process.cwd()` or something like that?

On that note, maybe these should have description headers comments?

mdonnalley: @amphro Is there a specific reason that you don't want `process.cwd()` to show up in help?

```~/code/plugin-login [main] : bin/dev login org jwt --help
Log in to your Salesforce orgs using a JSON web token

USAGE
  $ sf login org jwt

OPTIONS
  -a, --alias=<value>         Set an alias for the account or environment
  -d, --set-default           Set the org as the default org after login
  -f, --jwt-key-file=<value>  [default: /Users/mdonnalley/code/plugin-login] Path to a file containing the private key
  -i, --clientid=<value>      OAuth client ID (sometimes called the consumer key)
  -l, --instance-url=<value>  [default: https://login.salesforce.com] The login url
  -u, --username=<value>      Authentication username

EXAMPLE
  sf login org jwt --jwt-key-file myorg.key --username me@salesforce.com --clientid XXXX
```

I personally think it's pretty cool that it shows me the exact path it'll use as the default

src/parser/flags.ts
```diff
@@ -40,6 +40,34 @@ export const integer = build({
...
+  },
+})
+
+/**
+ * Validates that a string does not contain any invalid path characters.
+ * Defaults to the current working directory.
+ */
+export const path = build({
+  parse: async input => {
+    const invalidPathChars = /[["?<>|\]]+/
```
RodEsp, line 63: `\` is a valid directory separator for Windows

RodEsp (changes requested):

src/parser/flags.ts
amphro: You are reading that wrong. It is escaping `]` so it doesn't close the group.
```
> invalidPathChars.test('someDir]')
true // invalid
> invalidPathChars.test('someDir\dsadf')
false // valid
> invalidPathChars.test('someDir\\dsadf')
false // valid
```

RodEsp: Ooooohhhh, duh 🤦🏼 
Thanks.

RodEsp (approved):

src/parser/flags.ts
amphro: Because help is cached in `oclif.manifest`. So the end result is it will show the path of whatever machine generated the `oclif.manifest`, which is usually the build machine.

mdonnalley: Oh I see. Okay, I'll fix that up

PR #184, opened 2021-06-10
feat: help improvements and customizability

amphro:
### Command and Flag Summary
Oclif will use the first line of `command.description` as the "summary" which is used as the first line on command help as well as in the command list. This is clever, but can also be hidden and confusing. There is now a `summary` property on command and flag, which if provided, will be used as the first line on command help and in the command list. It will also be repeated in the `DESCRIPTION` section of help for increased usability.

The summary property should be one line.

### Flag Descriptions
Multiline flag descriptions can be hard to read in a wrapped list view. Oclif will render the list different if there is one descriptions is more than 4 lines, but at that point it may be too late. Not only that, it can still be useful to have a short, easy to digest flag list, with more description down below. Now the `summary` property on flag, if provided, will be used in the in the flag list and the description will be put in an extended section down below.

```
FLAGS
  -e, --target-env=<value>...  [default: dsfr] Environment name or alias to open.
  ...

DESCRIPTION
   ...

EXAMPLES
  ...

FLAG DESCRIPTIONS
  -e, --target-env=<value>...

    Specify the login user or alias that’s associated with the environment. For scratch orgs, the login user is generated
    by the command that created the scratch org. You can also set an alias for the scratch org when you create it.

    For Dev Hubs, sandboxes, and production orgs, specify the alias you set when you logged into the org with "sf login".
```

### Flag Help Groups
Help for commands with a large number of flags can be hard to read. It can be hard to pick out related flags. Or it can be hard to pick out flags that may not be directly related to the command, such as flags added by a CLI or a particular plugin. For example, `--json` or [table flags](https://github.com/oclif/cli-ux/#clitable). 

Flag help groups help solve this. 

```
FLAGS
  --[no-]opt  my boolean optional flag

TABLE FLAGS
  -x, --extended          show extra columns
  --columns=columns       only show provided columns (comma-separated)
  --csv                   output is csv format [alias: --output=csv]
  --filter=filter         filter property by partial string matching, ex: name=foo
  --hidden                show hidden commands
  --no-header             hide table header from output
  --no-truncate           do not truncate output to fit screen
  --output=csv|json|yaml  output in a more machine friendly format
  --sort=sort             property to sort by (prepend '-' for descending)

CLI FLAGS
  --json  format output as json
```

### Refactor help and extend customizability
There are time where a CLI may just want to add an extra help section or change a section heading in the command help. Or maybe the CLI wants to override the help options. With oclif@1, a CLI could really only override the `Help` class which did not provide any helpful methods for command help or options. Now, the `CommandHelp` class is loaded from the `Help` class so it can be extended. Also, help options can be defined in `pjson.oclif.helpOptions`. See `test/help/show-customized-help.test.ts`.

In addition, code like `wrap(body, this.opts.maxWidth - spacing, {hard: true, trim: false}` and creating a section with `[bold(header), indent(body, 2)].join('\n')` was called all over the place. These have been abstracted into a base class which all help classes now extend, providing those utilities to custom help classes as well.

### Example Descriptions
Having a list of examples with no descriptions is not very useful, which the existing example structure encourages. Consumers can add descriptions themselves but it is less than idea.

```javascript
Flag.string({
  description: 'my flag',
  examples: [
    'happy path\n<%= config.bin %> <%= command.id %> --flag'
    'run with foo bar\n<%= config.bin %> <%= command.id %> --flag --foo=bar'
  ]
})
```

Now, examples can be an object `{description, command}` which will be formatted appropriately. If a sting is provided, it will try to figure our what is a description and what is a command to format in a more readable output.
```
EXAMPLES
  happy path:

    $ bin command --flag

  run with foo bar:

    $ bin command --flag --foo=bar
```


src/help/formatter.ts
```diff
@@ -0,0 +1,176 @@
...
+   * ```
+   * @param body the text to indent
+   * @param spacing the final number of spaces this text will be indented
+   * @return the formatted indented text
+   */
+  public indent(body: string, spacing = this.indentSpacing): string {
+    return indent(body, spacing)
+  }
+
+  public renderList(input: (string | undefined)[][], opts: {indentation: number; multiline?: boolean; stripAnsi?: boolean; spacer?: string}): string {
```
amphro, line 105: This method is unchanged, just brought over from list.ts.

src/help/formatter.ts
```diff
@@ -0,0 +1,176 @@
...
+
+export class HelpFormatter {
+  indentSpacing = 2
+
+  /**
+   * Takes a string and replaces `<%= prop =>` with the value of prop, where prop is anything on
+   * `config=Interfaces.Config` or `opts=Interface.HelpOptions`.
+   *
+   * ```javascript
+   * '<%= config.bin =>` // will resolve to the bin defined in `pjson.oclif`.
```
mdonnalley, line 25: I think you meant to have a ` here instead of a '

src/command.ts
```diff
@@ -63,8 +69,21 @@ export default abstract class Command {
 
   static plugin: Interfaces.Plugin | undefined
 
-  /** An array of example strings to show at the end of the command's help */
-  static examples: string[] | undefined
+  /**
+   * An array of examples to show at the end of the command's help.
+   *
+   * IF only a string is provide, it will try to look for a line that starts
```
mdonnalley, line 75: Is this worth mentioning in the migration doc?

mdonnalley (approved):

mdonnalley (changes requested):
1. Found one issue while testing:

In the messages.md folder I have:
```markdown
# examples

- happy path example\n<%= config.bin %> <%= command.id %> --jwt-key-file myorg.key --username me@salesforce.com --clientid XXXX
```

But the help shows:
```
EXAMPLES
  happy path example\nsf login org jwt --jwt-key-file myorg.key --username me@salesforce.com --clientid XXXX
```
2. I noticed that the `FLAG DESCRIPTIONS` didn't behave as I expected: When I added a `summary` to one of my flags, I expected to see the new section in help but I didn't. I had to make sure that my `description` was also multiple lines in order to see the new section. I think my preference is to always add the `FLAG DESCRIPTIONS` section if the `summary` prop exists - regardless of whether or not my description is long enough
3. I think the `FLAGS` and `CLI FLAGS` distinction is unclear. I feel like there's a better way to communicate "these flags are specific to this command, and these other flags are generally available on all commands"
4. Lastly, if I understand correctly consumers have to implement flag groups on their own? Would it better to simply expose that as part of the flag options, e.g. 
```typescript
    alias: Flags.string({
      description: messages.getMessage('alias'),
      group: 'foo',
    }),
    clientid: Flags.string({
      description: messages.getMessage('clientId'),
      group: 'bar',
    }),
```

amphro:
1. You can't have `\n` in markdown since markdown is just text. Just hit enter, or if you want to use `\n` use a js or ts file instead. 
2. I agree. Fixed and added a test. 
3. Suggestions? I changed to "GLOBAL FLAGS" but I'm impartial. 
4. That is exactly how it is done. See how the --json flag 
```
  alias: Flags.string({
      description: messages.getMessage('alias'),
      helpGroup: 'foo',
  }),
  clientid: Flags.string({
      description: messages.getMessage('clientId'),
      helpGroup: 'bar',
  }),
```

src/help/command.ts
```diff
@@ -69,45 +133,87 @@ export default class CommandHelp {
...
-    if (!description) return
-    return [
-      bold('DESCRIPTION'),
-      indent(wrap(description.trim(), this.opts.maxWidth - 2, {trim: false, hard: true}), 2),
-    ].join('\n')
+
+    let description: string[]
+
+    if (this.opts.hideCommandSummaryInDescription) {
+      description = (cmd.description || '').split('\n').slice(1)
```
RodEsp, line 140:
```suggestion
      description = (cmd.description || '').split(EOL).slice(1)
```

src/help/command.ts
```diff
@@ -69,45 +133,87 @@ export default class CommandHelp {
...
+    } else {
+      description = [
+        ...(cmd.summary || '').split('\n'),
+        ...(cmd.description || '').split('\n'),
```
RodEsp, line 144:
```suggestion
        ...(cmd.summary || '').split(EOL),
        ...(cmd.description || '').split(EOL),
```

RodEsp (commented):
Consider changing all the references to `\n` for `EOL` for windows compatibility?

src/help/index.ts
```diff
@@ -127,8 +114,8 @@ export class Help extends HelpBase {
     const subTopics = this.sortedTopics.filter(t => t.name.startsWith(name + ':') && t.name.split(':').length === depth + 1)
     const subCommands = this.sortedCommands.filter(c => c.id.startsWith(name + ':') && c.id.split(':').length === depth + 1)
 
-    const title = command.description && this.render(command.description).split('\n')[0]
-    if (title) console.log(title + '\n')
+    const summary = this.summary(command) // command.description && this.render(command.description).split('\n')[0]
+    if (summary) console.log(summary + '\n')
```
RodEsp, line 118: This is still pulling from the command description, not the `summary` but I can't understand why.

src/help/index.ts
```diff
@@ -197,55 +184,60 @@ export class Help extends HelpBase {
...
-      bold('COMMANDS'),
-      indent(body, 2),
-    ].join('\n')
+    return this.section('COMMANDS', body)
+  }
+
+  protected summary(c: Interfaces.Command): string {
+    if (c.summary) return this.render(c.summary.split('\n')[0])
+
+    return this.render((c.description || '').split('\n')[0])
```
RodEsp, line 212: I believe this should be `c.summary` but for some reason the `.summary` isn't being added to the command. See comment on [line 118](https://github.com/oclif/core/pull/184/files#diff-ee5d33646fe2e22a534294dd09bb1a83b79742e98b3dc681175fe5754427b5f9R118)
```suggestion
    return this.render((c.summary|| '').split('\n')[0])
```

RodEsp (changes requested):

RodEsp (changes requested):
Super long flag descriptions don't align correctly on Windows. They just continue on a newline at the same indentation as the flags.
image

src/help/index.ts
RodEsp: I don't know if this is related but `.\bin\dev commandreference generate` seems to not be able to find `.name` [in here](https://github.com/forcedotcom/plugin-command-reference/blob/d7b6ef630cb383bbd963cb095f73c0a0ad1c1590/src/docs.ts#L145) when this branch is `yarn link`ed.

So it's like it's not getting the `.plugin` property [here](https://github.com/oclif/core/pull/184/files#diff-1212ed35aabbfc158720210527b07c5e3ccaf889121fa334acc54b160afe561cR70).

src/help/index.ts
RodEsp: This test exemplifies what I mean
https://github.com/oclif/core/pull/186

src/help/command.ts
```diff
@@ -282,7 +282,15 @@ export class CommandHelp extends HelpFormatter {
     if (flagsWithExtendedDescriptions.length === 0) return
 
     const body = flagsWithExtendedDescriptions.map(flag => {
-      return `${this.flagHelpLabel(flag, true)}  ${flag.summary}\n\n${this.indent(this.wrap(flag.description || '', this.indentSpacing * 2))}`
+      // Guaranteed to be set because of the filter above, but make ts happy
```
RodEsp, line 285: Ugh, lol.

RodEsp (approved):

PR #187, opened 2021-06-28
feat: return results from runHook

mdonnalley:


peternhale (approved):
@mdonnalley just curious if the hooks can be run concurrently?

PR #202, opened 2021-07-13
feat: add DAG class

peternhale:
@mdonnalley @RodEsp just wanted to get the graph impl reviewed as I start the integration with this work and config.

DO NOT MERGE

### DIFF
--- a/src/config/graph.ts
+++ b/src/config/graph.ts
@@ -1,0 +1,138 @@
+enum VisitedColor {
+  NONE,
+  RED,
+  GREEN,
+  BLACK,
+  BLUE
+}
+export interface Node {
+  id: string;
+  color?: VisitedColor;
+  visitCnt?: number;
+}
+
+export interface Edge {
+  source: Node;
+  sink: Node;
+}
+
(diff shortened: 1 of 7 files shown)


mdonnalley (commented):
Looks great so far - awesome work!

src/config/graph.ts
```diff
@@ -0,0 +1,139 @@
...
+  setNode(name: string, node: N): void {
+    this._nodes.set(name, node)
+  }
+
+  setEdge(source: N, sink: N): EdgeClass<N, E> {
+    this.setNode(source.id, source)
+    this.setNode(sink.id, sink)
+
+    const edge = new EdgeClass<N, E>(source, sink)
+    if (this._edges.has(edge)) {
```
amphro, line 49: Will this every return true? `edge` is a new edge instance which won't be in the set. I think you want something like `this._edges.find(edge => edge.source.id === source.id && edge.sink.id === sink.id)`?

peternhale: I will check

peternhale: Nope, always creates a new edge. Nice catch.

PR #205, opened 2021-07-26
fix: restore short flags for --help and --version

peternhale:
@W-9612538@

### DIFF
--- a/src/flags.ts
+++ b/src/flags.ts
@@ -30,7 +30,7 @@ export {boolean, integer, url} from './parser'
 
 export const version = (opts: Partial<BooleanFlag<boolean>> = {}) => {
   return Parser.flags.boolean({
-    // char: 'v',
+    char: 'v',
     description: 'show CLI version',
     ...opts,
     parse: async (_: any, cmd: Command) => {
@@ -42,7 +42,7 @@ export const version = (opts: Partial<BooleanFlag<boolean>> = {}) => {
 
 export const help = (opts: Partial<BooleanFlag<boolean>> = {}) => {
   return Parser.flags.boolean({
-    // char: 'h',
+    char: 'h',
     description: 'show CLI help',
     ...opts,
     parse: async (_: any, cmd: Command) => {
--- a/src/flags.ts
+++ b/src/flags.ts
@@ -30,7 +30,6 @@ export {boolean, integer, url} from './parser'
 
 export const version = (opts: Partial<BooleanFlag<boolean>> = {}) => {
   return Parser.flags.boolean({
-    char: 'v',
     description: 'show CLI version',
     ...opts,
     parse: async (_: any, cmd: Command) => {
@@ -42,7 +41,6 @@ export const version = (opts: Partial<BooleanFlag<boolean>> = {}) => {
 
 export const help = (opts: Partial<BooleanFlag<boolean>> = {}) => {
   return Parser.flags.boolean({
-    char: 'h',
     description: 'show CLI help',
     ...opts,
     parse: async (_: any, cmd: Command) => {
--- a/src/interfaces/pjson.ts
+++ b/src/interfaces/pjson.ts
@@ -43,8 +43,8 @@ export namespace PJSON {
           hidden?: boolean;
         };
       };
-      helpOverride?: string[];
-      versionOverride?: string[];
+      additionalHelpFlags?: string[];
+      additionalVersionFlags?: string[];
     };
   }
 
--- a/src/interfaces/pjson.ts
+++ b/src/interfaces/pjson.ts
@@ -43,6 +43,8 @@ export namespace PJSON {
           hidden?: boolean;
         };
       };
+      helpOverride?: string[];
+      versionOverride?: string[];
     };
   }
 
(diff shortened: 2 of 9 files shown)


mdonnalley (approved):

src/flags.ts
```diff
@@ -42,7 +42,7 @@ export const version = (opts: Partial<BooleanFlag<boolean>> = {}) => {
 
 export const help = (opts: Partial<BooleanFlag<boolean>> = {}) => {
   return Parser.flags.boolean({
-    // char: 'h',
+    char: 'h',
```
mdonnalley, line 45: Do we still want this change since it's now configurable?

src/interfaces/pjson.ts
```diff
@@ -43,6 +43,8 @@ export namespace PJSON {
           hidden?: boolean;
         };
       };
+      helpOverride?: string[];
```
mdonnalley, line 46: I'm not sure that `helpOverride` and `versionOverride` are the best words here. We're not really overriding any behavior but instead adding to existing behavior.

Maybe it'd make more sense to make this `helpFlags` and `versionFlags`? 

As I'm thinking about it now, how useful is it to provide this functionality? Are users really going to use anything other than `-h` and `-v` for help and version? So perhaps it'd be better to have a couple of boolean properties that toggle on the short flags, e.g. `enableHelpShortFlag` and `enableVersionShortFlag`

I'm not sure what the right answer, I'm just thinking out loud here

src/flags.ts
peternhale: Do not need this change, will revert.

src/interfaces/pjson.ts
peternhale: The name thing bothered me too. I'll think about new names.

For sfdx version we have `-v`, `--version` and `version`, so we have opened the door for flag names that are outside the short names. Core also supports the notion of command patterns like `<cli> --help` or `<cli> help` or `<cli> <cmd> [--help|help]` or `<cli> [--help|help] <cmd>`. Having a list of additional flags allows one to specify any name for "help."

IMO, I see us as the only ones going to use this capability, others will probably adapt to core's defaults.

mdonnalley: `help` and `version` are commands though, not flags. The `help` command is provided by [plugin-help](https://github.com/oclif/plugin-help) and the `--help` flag is provided by core. So if we're going to offer a `version` command I think we ought to make a plugin-version to be consistent with the existing pattern

I'm okay with providing the ability to configure any flag name but I agree that we might be the only ones who actually use it.

peternhale: These flags are intercepted in main.ts, with each having its own block [help](https://github.com/oclif/core/blob/phale/W-9612538/src/main.ts#L53) and [version](https://github.com/oclif/core/blob/phale/W-9612538/src/main.ts#L47). When a version flag or version command is detected it delegates to `config.userAgent`. When running `sf help deploy`, that is trapped by the block at line 53 and then delegated to a Help class. The only time I see main.ts delegate to the command, is when topic is used w/o a help flag.

mdonnalley: @peternhale I'm not sure that's correct. The debug logs indicate that `sf help deploy` is delegated to plugin-help:

```
  config runCommand help [ 'deploy' ] +1ms
  config:@oclif/plugin-help (require) /Users/mdonnalley/.nvm/versions/node/v14.15.4/lib/node_modules/@salesforce/cli/node_modules/@oclif/plugin-help/lib/commands/help.js +5ms
/Users/mdonnalley/.nvm/versions/node/v14.15.4/lib/node_modules/@salesforce/cli/node_modules/@sf/telemetry/lib/hooks/telemetryPrerun.js +0ms
  sf:help init version: @oclif/command@1.8.0 argv: [ 'deploy' ] +0ms
```

I don't think any additional works needs to be done in this PR. My point is just that if we're going to add a `version` command, it would be more consistent to place the functionality in a plugin.

peternhale: @mdonnalley I will compare current behavior wilh my changes to see if I have introduced an unwanted side ef
fect

mdonnalley (approved):

test/config/help.config.test.ts
```diff
@@ -0,0 +1,51 @@
...
+const root = path.resolve(__dirname, 'fixtures/help')
+// const p = (p: string) => path.join(root, p)
+
+// This tests file URL / import.meta.url simulation.
+const rootAsFileURL = url.pathToFileURL(root).toString()
+
+const withConfig = fancy
+.add('config', () => Config.load(rootAsFileURL))
+
+describe('esm', () => {
```
RodEsp, line 19: What's `esm`?

RodEsp (changes requested):
I second Mike's comments, https://github.com/oclif/core/pull/205#discussion_r677539916.

I also see that the config properties were changed from "override" to "additional" but it seems like everything in the code itself is still using the "override" verbiage. I'd like to see that changed too for consistency and readability.

test/config/help.config.test.ts
peternhale: copy/paste error, but core tests different esm module types. Will enter an appropriate description

peternhale:
> I second Mike's comments, [#205 (comment)](https://github.com/oclif/core/pull/205#discussion_r677539916). Per the `sf` styleguide and generally accepted CLI standards if something has `--` at the front it's a long flag, if it has `-` it's a short flag, if it doesn't have either it's a command.
> 
> I also see that the config properties were changed from "override" to "additional" but it seems like everything in the code itself is still using the "override" verbiage. I'd like to see that changed too for consistency and readability.

I have made the adjustments that protect cmds help and version. Fixing refs to override to be more inline with context.

RodEsp (approved):

PR #206, opened 2021-07-30
feat: add state property

mdonnalley:
Add a `state` property to the `Command` and `Config` classes to support marking commands and/or plugins as beta

### DIFF
--- a/src/help/index.ts
+++ b/src/help/index.ts
@@ -39,14 +39,14 @@ export abstract class HelpBase extends HelpFormatter {
    * Show help, used in multi-command CLIs
    * @param args passed into your command, useful for determining which type of help to display
    */
-  public abstract async showHelp(argv: string[]): Promise<void>;
+  public abstract showHelp(argv: string[]): Promise<void>;
 
   /**
    * Show help for an individual command
    * @param command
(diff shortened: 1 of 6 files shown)


src/help/index.ts
```diff
@@ -122,6 +122,10 @@ export class Help extends HelpBase {
 
     const subTopics = this.sortedTopics.filter(t => t.name.startsWith(name + ':') && t.name.split(':').length === depth + 1)
     const subCommands = this.sortedCommands.filter(c => c.id.startsWith(name + ':') && c.id.split(':').length === depth + 1)
+    const plugin = this.config.plugins.find(p => p.name === command.pluginName)
+    console.log(this.config.pjson?.oclif?.state, plugin?.pjson?.oclif?.state, command.state)
```
peternhale, line 126: @mdonnalley Please remove if not needed

peternhale (changes requested):

RodEsp (approved):
I have a feeling we'll need to change the `This command is in ${state}` verbiage but that'll come after design decisions.

Looks good!

peternhale (approved):

PR #231, opened 2021-08-18
fix: command name parsing when flag=value present

mdonnalley:
[skip-validate-pr]

### DIFF
--- a/src/help/util.ts
+++ b/src/help/util.ts
@@ -43,7 +43,7 @@ function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
   const findId = (argv: string[]): string | undefined => {
     const final: string[] = []
     const idPresent = (id: string) => ids.includes(id)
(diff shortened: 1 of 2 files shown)


src/help/util.ts
```diff
@@ -43,7 +43,7 @@ function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
   const findId = (argv: string[]): string | undefined => {
     const final: string[] = []
     const idPresent = (id: string) => ids.includes(id)
-    const isFlag = (s: string) => s.startsWith('-')
+    const isFlag = (s: string) => s.startsWith('-') || s.includes('=')
```
peternhale, line 46: @mdonnalley not sure this is the right place to do this. It catches var args with the `key=value` pattern, but not simple args.

peternhale (changes requested):

peternhale (approved):

RodEsp (approved):

PR #233, opened 2021-08-24
fix: initial changes to add error codes

peternhale:
@W-9791511@

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -9,8 +9,6 @@ import * as Errors from './errors'
 import {PrettyPrintableError} from './errors'
 import * as Parser from './parser'
 import * as Flags from './flags'
-import {HelpSection} from './help/command'
-import {ErrorCode} from './interfaces/command'
 
 const pjson = require('../package.json')
 
@@ -93,8 +91,6 @@ export default abstract class Command {
   static parserOptions = {}
 
   static disableJsonFlag: boolean | undefined
-
-  static errorCodes?: ErrorCode[];
 
   // eslint-disable-next-line valid-jsdoc
   /**
--- a/src/command.ts
+++ b/src/command.ts
@@ -92,6 +92,17 @@ export default abstract class Command {
 
   static disableJsonFlag: boolean | undefined
 
+  /**
+   * Provides a list of command properties that need to be included in the manifest file, which then makes the
+   * properties available to command help.
+   */
+  additionalPropertiesForManifest?: string[];
+
+  /**
+   * Remaining properties defined by developer
+   */
+  [name: string]: unknown;
+
   // eslint-disable-next-line valid-jsdoc
   /**
    * instantiate and run the command
--- a/src/command.ts
+++ b/src/command.ts
@@ -9,6 +9,8 @@ import * as Errors from './errors'
 import {PrettyPrintableError} from './errors'
 import * as Parser from './parser'
 import * as Flags from './flags'
+import {HelpSection} from './help/command'
+import {ErrorCode} from './interfaces/command'
 
 const pjson = require('../package.json')
 
@@ -91,6 +93,8 @@ export default abstract class Command {
   static parserOptions = {}
 
   static disableJsonFlag: boolean | undefined
+
+  static errorCodes?: ErrorCode[];
 
   // eslint-disable-next-line valid-jsdoc
   /**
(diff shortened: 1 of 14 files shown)


src/command.ts
```diff
@@ -92,6 +94,8 @@ export default abstract class Command {
 
   static disableJsonFlag: boolean | undefined
 
+  static errorCodes?: ErrorCode[];
```
peternhale, line 97: Add new static property for command to hold user defined error codes

src/help/command.ts
```diff
@@ -119,10 +119,17 @@ export class CommandHelp extends HelpFormatter {
         header: 'FLAG DESCRIPTIONS',
         generate: ({flags}) => this.flagsDescriptions(flags),
       },
+      {
```
peternhale, line 122: Account for new help section, `ERROR CODES`, in default CommandHelp class.

src/help/command.ts
```diff
@@ -311,5 +318,10 @@ export class CommandHelp extends HelpFormatter {
 
     return body
   }
+
+  public errorCodes(errorCodes: ErrorCode[]): string | undefined {
```
peternhale, line 322: format error codes for display in help

src/help/index.ts
```diff
@@ -50,7 +50,7 @@ export abstract class HelpBase extends HelpFormatter {
 }
 
 export class Help extends HelpBase {
-  protected CommandHelpClass: typeof CommandHelp = CommandHelp
+  public CommandHelpClass: typeof CommandHelp = CommandHelp
```
peternhale, line 53: Modified visibility to allow one to override this value

src/index.ts
```diff
@@ -7,22 +7,26 @@ import {Config, Plugin, tsPath, toCached} from './config'
...
+import {CommandHelp, HelpBase, Help, loadHelpClass} from './help'
 import {toStandardizedId, toConfiguredId} from './help/util'
 import * as Parser from './parser'
 import {Hook} from './interfaces/hooks'
 import {settings, Settings} from './settings'
+import {HelpSection, HelpSectionRenderer} from './help/command'
 
 const flush = require('../flush')
 
 export {
```
peternhale, line 19: Export additional elements needed  by custom help classes

src/interfaces/command.ts
```diff
@@ -6,6 +6,11 @@ export type Example = string | {
   command: string;
 }
 
+export type ErrorCode = {
```
peternhale, line 9: Type for error codes

src/command.ts
mdonnalley: The more I think about it, the more I think this should belong in our custom help class instead of oclif

My main concern is not overwhelming oclif users with all the possible options

src/help/index.ts
mdonnalley: What issues were you running into with overriding protected methods? That's something you should be able to do. 

I tried the following and typescript didn't complain about it:

```typescript
export class MyCommandHelp extends CommandHelp {
  protected sections(): Array<{header: string; generate: HelpSectionRenderer }> {
    const generate = (() => 'a string') as HelpSectionRenderer;
    return [{ header: 'my section', generate }];
  }
}

export class MyHelp extends Help {
  protected CommandHelpClass: typeof MyCommandHelp = MyCommandHelp
}
```

src/command.ts
peternhale: I will move error codes elsewhere

src/help/formatter.ts
```diff
@@ -163,13 +167,31 @@ export class HelpFormatter {
...
+    } else if (Array.isArray(body)) {
+      newBody = (body! as [string, string | undefined | HelpSectionTable][]).map(entry => {
+        if (Reflect.has(entry, 'name')) {
+          const tableEntry = entry as unknown as {name: string; description: string}
+          return ([this.render(tableEntry.name), this.render(tableEntry.description)])
+        }
+        const [left, right] = entry
+        return ([this.render(left), right && this.render(right as string)])
+      })
+    } else if (Reflect.has(body!, 'header')) {
```
petehale, line 184: replace with 'x' in y

mdonnalley (approved):

RodEsp (changes requested):

RodEsp (approved):

PR #244, opened 2021-09-13
fix: move ctor for command help class to its own function

peternhale:
@W-9708749@

### DIFF
--- a/src/help/index.ts
+++ b/src/help/index.ts
@@ -30,19 +30,9 @@ function getHelpSubject(args: string[], config: Interfaces.Config): string | und
 }
 
 export abstract class HelpBase extends HelpFormatter {
-  private argvCopy: string[] = []
-
   constructor(config: Interfaces.Config, opts: Partial<Interfaces.HelpOptions> = {}) {
     super(config, opts)
     if (!config.topicSeparator) config.topicSeparator = ':' // back-support @oclif/config
-  }
-
-  public get originalArgv(): string[] {
-    return this.argvCopy
-  }
-
-  public set originalArgv(argv: string[]) {
-    this.argvCopy = argv
   }
 
   /**
@@ -100,6 +90,8 @@ export class Help extends HelpBase {
   }
 
   public async showHelp(argv: string[]) {
+    argv = argv.filter(arg => !getHelpFlagAdditions(this.config).includes(arg))
+
     if (this.config.topicSeparator !== ':') argv = standardizeIDFromArgv(argv, this.config)
     const subject = getHelpSubject(argv, this.config)
     if (!subject) {
--- a/src/help/index.ts
+++ b/src/help/index.ts
@@ -30,9 +30,19 @@ function getHelpSubject(args: string[], config: Interfaces.Config): string | und
 }
 
 export abstract class HelpBase extends HelpFormatter {
+  private argvCopy: string[] = []
+
   constructor(config: Interfaces.Config, opts: Partial<Interfaces.HelpOptions> = {}) {
     super(config, opts)
     if (!config.topicSeparator) config.topicSeparator = ':' // back-support @oclif/config
+  }
+
+  public get originalArgv(): string[] {
+    return this.argvCopy
+  }
+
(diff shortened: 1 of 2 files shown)


mdonnalley (approved):

src/main.ts
```diff
@@ -54,6 +55,7 @@ export async function run(argv = process.argv.slice(2), options?: Interfaces.Loa
     argv = argv.filter(arg => !getHelpFlagAdditions(config).includes(arg))
     const Help = await loadHelpClass(config)
     const help = new Help(config, config.pjson.helpOptions)
```
peternhale, line 57: I did not want to add another parameter to the constructor, so capturing original argv and then setting it directly o n the Help class instance.

src/main.ts
```diff
@@ -30,6 +30,7 @@ export const versionAddition = (argv: string[], config?: Interfaces.Config): boo
 }
 
 export async function run(argv = process.argv.slice(2), options?: Interfaces.LoadOptions) {
+  const originalArgv = [...argv]
```
peternhale, line 33: Capture argv as passed to run

src/main.ts
```diff
@@ -54,6 +55,7 @@ export async function run(argv = process.argv.slice(2), options?: Interfaces.Loa
     argv = argv.filter(arg => !getHelpFlagAdditions(config).includes(arg))
     const Help = await loadHelpClass(config)
     const help = new Help(config, config.pjson.helpOptions)
+    help.originalArgv = originalArgv
```
peternhale, line 58: set the value on the help class.

src/help/index.ts
```diff
@@ -30,11 +30,21 @@ function getHelpSubject(args: string[], config: Interfaces.Config): string | und
...
 
 export abstract class HelpBase extends HelpFormatter {
+  private argvCopy: string[] = []
+
   constructor(config: Interfaces.Config, opts: Partial<Interfaces.HelpOptions> = {}) {
     super(config, opts)
     if (!config.topicSeparator) config.topicSeparator = ':' // back-support @oclif/config
   }
 
+  public get originalArgv(): string[] {
```
peternhale, line 40: getter/setter

RodEsp:
Review comments captured in slack: https://platformcloud.slack.com/archives/C01U75RNU1G/p1631567124234900?thread_ts=1631553503.228700&cid=C01U75RNU1G

src/help/index.ts
```diff
@@ -90,6 +90,8 @@ export class Help extends HelpBase {
   }
 
   public async showHelp(argv: string[]) {
+    argv = argv.filter(arg => !getHelpFlagAdditions(this.config).includes(arg))
```
peternhale, line 93: Moved the stripping of help flags to showHelp from main.ts.

Other function call via showHelp expect the help flags to be removed.

src/main.ts
```diff
@@ -51,7 +51,6 @@ export async function run(argv = process.argv.slice(2), options?: Interfaces.Loa
 
   // display help version if applicable
   if (helpAddition(argv, config)) {
-    argv = argv.filter(arg => !getHelpFlagAdditions(config).includes(arg))
```
peternhale, line 54: main.ts should not have any knowledge of what showHelp expects in argv, just pass the whole thing as is.

src/help/index.ts
```diff
@@ -202,10 +204,14 @@ export class Help extends HelpBase {
       command.id = command.id.replace(/:/g, this.config.topicSeparator)
       command.aliases = command.aliases && command.aliases.map(a => a.replace(/:/g, this.config.topicSeparator))
     }
-    const help = new this.CommandHelpClass(command, this.config, this.opts)
+    const help = this.getCommandHelpClass(command)
```
RodEsp, line 207: Do we really need a whole new method for this if this is the only place it's used?

RodEsp (approved):

src/help/index.ts
peternhale: @RodEsp This function will be overridden in the custom help class to allow the determination of short help.

PR #253, opened 2021-09-17
feat: parallelize runHook

mdonnalley:
Use `Promise.all` to parallelize hooks

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -225,10 +225,11 @@ export class Config implements IConfig {
       })
     }
 
-    const successes = []
-    const failures = []
-
-    for (const p of this.plugins) {
+    const final = {
+      successes: [],
+      failures: [],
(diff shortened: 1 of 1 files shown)


src/config/config.ts
```diff
@@ -225,10 +225,11 @@ export class Config implements IConfig {
...
 
-    const successes = []
-    const failures = []
-
-    for (const p of this.plugins) {
+    const final = {
+      successes: [],
+      failures: [],
+    } as Hook.Result<Hooks[T]['return']>
+    const promises = this.plugins.map(async p => {
```
RodEsp, line 232:
```suggestion
    } as Hook.Result<Hooks[T]['return']>

    const promises = this.plugins.map(async p => {
```

RodEsp (approved):

PR #258, opened 2021-09-25
Replace Chalk with Colorette

NickKaramoff:
[Colorette][1] is a very fast and lightweight color formatting library. It is faster than alternatives in both loading and calling times and saves precious space in `node_modules`.

After doing the same operation in cli-ux (see https://github.com/oclif/cli-ux/pull/418) as well as oclif/parser, the overall dependency tree of oclif will be less clunky, and the install size will be smaller.

P. S. I have opened https://github.com/oclif/core/pull/257 before this PR, but then I replaced Nano Colors with Colorette in order not to take part in a [weird situation around the creation of the former][2]

[1]: https://github.com/jorgebucaran/colorette
[2]: https://twitter.com/jorgebucaran/status/1441713235980275718


RodEsp:
Hey @NickKaramoff, thanks for the PR but chalk has a way larger community than colorette so for the sake of stability and "support" we won't be switching over for now.

NickKaramoff:
> chalk has a way larger community than colorette so for the sake of stability and "support" we won't be switching over for now.

Terminal styling is done using ANSI sequences, so, for best support and maintainability, a colouring library should just be a bunch of strings, which Chalk definitely isn't. Colorette (or picocolors) may not have the clout of Chalk, but they have the source code that everyone is able to understand and no extraneous and/or vendored dependencies. Not only is it easier to detect and fix vulnerabilities in packages like these, it's almost impossible for them to even have a vulnerability.

But alas, it's your project and your decisions, and I won't argue much against it 😅

PR #290, opened 2021-11-18
fix: resolve typescript compilation errors

mdonnalley:
Fixes issues caused by upgrading to typescript@4.5.2

@W-10190107@

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@oclif/core",
   "description": "base library for oclif CLIs",
   "version": "1.0.3",
-  "author": "Jeff Dickey @jdxcode",
+  "author": "Salesforce",
   "bugs": "https://github.com/oclif/core/issues",
   "dependencies": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@oclif/core",
   "description": "base library for oclif CLIs",
   "version": "1.0.3",
-  "author": "Jeff Dickey @jdxcode",
+  "author": "Salesforce",
```
RodEsp, line 5: Maybe `Salesforce CLI Team`?
Doesn't really matter tho.

RodEsp (approved):

package.json
mdonnalley: I just went with what's in sfdx and sf

https://github.com/salesforcecli/cli/blob/main/package.json#L5
https://github.com/salesforcecli/sfdx-cli/blob/main/package.json#L5

PR #294, opened 2021-11-24
feat: Add support for typescript files and ESModules.

lukasholzer:
When relying on third party dependencies that are [ESModules](https://nodejs.org/api/esm.html#modules-ecmascript-modules) (which is getting more popularity in the Node.js community it is important to be invoked by an ESModule. 

Currently the `@oclif/core` already works when using pure ESModules by setting the `"type": "module"` inside the `package.json`.

But if you want to leverage `typescript` you currently need to compile the typescript sources to the `lib` folder by running `npx tsc --project tsconfig.json`. This will compile the command inside the lib folder. After that running `./bin/run.js` will work.

IMHO this developer experience can be enhanced by leveraging ts-node to directly run the typescript sources.

All that the developer now needs to do is to specify the `ts-node/esm` loader when invoking the binary by updating the binary file with the following content:

The file `bin/run.js`

```javascript
#!/usr/bin/env node --experimental-specifier-resolution=node --no-warnings --loader ts-node/esm
import { Errors, flush, run } from '@oclif/core';

run(void 0, import.meta.url)
  .then(flush)
  .catch(Errors.handle);
```

now it can be invoked by running `NODE_ENV=development ./bin/run.js` and it will use `ts-node` to transpile the modules on the fly and resolve them correctly with this changes.

A better DX as first compiling and then running IMHO.

### DIFF
--- a/src/module-loader.ts
+++ b/src/module-loader.ts
@@ -7,13 +7,12 @@ import {Config as IConfig} from './interfaces'
 import {Plugin as IPlugin} from './interfaces'
 import * as Config from './config'
 
-const getPackageType = require('get-package-type')
 
 /**
  * Defines file extension resolution when source files do not have an extension.
  */
 // eslint-disable-next-line camelcase
-const s_EXTENSIONS: string[] = ['.js', '.mjs', '.cjs']
+const s_EXTENSIONS: string[] = ['.js', '.mjs', '.cjs', '.ts']
 
 /**
  * Provides a mechanism to use dynamic import / import() with tsconfig -> module: commonJS as otherwise import() gets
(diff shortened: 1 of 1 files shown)


lukasholzer:
I've signed the CLA but it seems not to update:

CleanShot 2021-11-24 at 13 11 54

src/module-loader.ts
```diff
@@ -99,17 +98,18 @@ export default class ModuleLoader {
...
-  static isPathModule(filePath: string): boolean {
+  static isPathModule(filePath: string, config: IConfig|IPlugin): boolean {
     const extension = path.extname(filePath).toLowerCase()
 
     switch (extension) {
+    case '.ts':
     case '.js':
-      return getPackageType.sync(filePath) === 'module'
-
+      return config.pjson?.type === 'module'
```
lukasholzer, line 112: No need to have the `get-package-type` as the rule is:

### "commonjs" Behavior
`.cjs` files are treated as CommonJS modules
`.js` files are treated as CommonJS modules
`.mjs` files are treated as ES modules

### "module" Behavior
`.cjs` files are treated as CommonJS modules
`.js` files are treated as ES modules
`.mjs` files are treated as ES modules

So the difference on how `.js` files are treated only relies on the type field inside the `package.json`
As you can see, the difference is how .js files are treated.

zipzapzanigan:
Yes please! I am trying to do exactly this with a library that is required to be an es module, and I am using typescript and I am pulling my hair out. +1 +1 for getting this pulled in

PR #307, opened 2021-12-01
feat: add table flags

mdonnalley:
Adds table flags formerly in [cli-ux](https://github.com/oclif/cli-ux/blob/master/src/styled/table.ts)

### Usage

```typescript
export default class EnvList extends Command {
  public static flags = {
    all: Flags.boolean({
      summary: messages.getMessage('flags.all.summary'),
      char: 'a',
    }),
    ...Flags.table(), // adds all table related flags
  };

}
```

@W-10255961@

### DIFF
--- a/src/flags.ts
+++ b/src/flags.ts
@@ -49,3 +49,20 @@ export const help = (opts: Partial<BooleanFlag<boolean>> = {}) => {
     },
   })
 }
+
+export const table = () => {
+  return {
+    columns: Parser.flags.string({exclusive: ['extended'], description: 'Only show provided columns (comma-separated).'}),
+    sort: Parser.flags.string({description: 'Property to sort by (prepend \'-\' for descending).'}),
+    filter: Parser.flags.string({description: 'Filter property by partial string matching, ex: name=foo.'}),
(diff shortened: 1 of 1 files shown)


peternhale (approved):

RodEsp:
Hmm... do we even need this? [EnvList doesn't use the flags from cli-ux](https://github.com/salesforcecli/plugin-env/blob/main/src/commands/env/list.ts#L43). I don't want to tightly couple this to the args that the cli-ux table takes.

mdonnalley:
> Hmm... do we even need this? [EnvList doesn't use the flags from cli-ux](https://github.com/salesforcecli/plugin-env/blob/main/src/commands/env/list.ts#L43). I don't want to tightly couple this to the args that the cli-ux table takes.

That's a good question, I was wondering the same... I think it's a nice thing to keep for the community. I'm assume someone out there is using it

RodEsp:
I'd prefer not to merge this cuz I can easily see becoming outdated and/or breaking if the cli-ux table options change but if you guys think it's better for the community this way then that's fine too. I'd just add a comment about how these need to reflect the [table options](https://github.com/oclif/cli-ux/blob/530606a6be7ee44192a154092146da31a5170fe9/src/styled/table.ts#L365-L375).

PR #322, opened 2021-12-13
Add missing await to README.md

gingermusketeer:
Noticed this while migrating to @oclif/core from @oclif/command.


RodEsp (approved):

RodEsp:
Good catch @gingermusketeer! Thanks!

PR #343, opened 2022-01-08
Fix flags cap

eob:
The `Flags` object exported from `@oclif/core` begins with a capital F (in prior versions it was a small `f`)


RodEsp (approved):

RodEsp:
Thanks for the contribution @eob!

PR #352, opened 2022-01-28
fix: module resolution of linked plugins

mdonnalley:
- Fixes module resolution for linked plugins
- Fixes issue with `command_not_found` hook always throwing an error - even if the correct command was run by `plugin-not-found`

@W-10501665@

### DIFF
--- a/src/config/plugin.ts
+++ b/src/config/plugin.ts
@@ -83,7 +83,7 @@ async function findRoot(name: string | undefined, root: string) {
   if (name) {
     let pkgPath
     try {
-      pkgPath = resolvePackage(name, {paths: [__dirname, root]})
+      pkgPath = resolvePackage(name, {paths: [root]})
     } catch {}
 
     return pkgPath ? findSourcesRoot(path.dirname(pkgPath)) : findRootLegacy(name, root)
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -281,7 +281,13 @@ export class Config implements IConfig {
     debug('runCommand %s %o', id, argv)
     const c = cachedCommand || this.findCommand(id)
     if (!c) {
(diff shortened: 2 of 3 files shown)


src/config/plugin.ts
```diff
@@ -83,7 +83,7 @@ async function findRoot(name: string | undefined, root: string) {
   if (name) {
     let pkgPath
     try {
-      pkgPath = resolvePackage(name, {paths: [__dirname, root]})
+      pkgPath = resolvePackage(name, {paths: [root]})
```
peternhale, line 86: @mdonnalley why is __dirname being removed?

peternhale (changes requested):

src/config/plugin.ts
mdonnalley: @peternhale two reasons: it's redundant and it was causing linked plugins to resolve to the version in node_modules - not your local version.

In the case of an unlinked plugin, we have
`__dirname` => `/Users/mdonnalley/.nvm/versions/node/v17.1.0/lib/node_modules/@salesforce/cli/node_modules/@oclif/core/lib/config`
`root` => `/Users/mdonnalley/.nvm/versions/node/v17.1.0/lib/node_modules/@salesforce/cli`

In this case, `__dirname` is redundant since it's already included by the `root` path

In the case of a linked plugin, we have

`__dirname` => `/Users/mdonnalley/.nvm/versions/node/v17.1.0/lib/node_modules/@salesforce/cli/node_modules/@oclif/core/lib/config`
`root` => `/Users/mdonnalley/repos/oclif/plugin-version`

In this case, `__dirname` was causing the resolution to return the node_modules version instead of the local version

In both cases, `root` is sufficient for resolving the plugin's location

peternhale (approved):

PR #354, opened 2022-02-01
feat: add S3Manifest

mdonnalley:
Adds `S3Manifest` interface which will be used by [oclif](https://github.com/oclif/oclif/blob/main/src/tarballs/config.ts#L33) and [plugin-update](https://github.com/oclif/plugin-update/blob/main/src/update.ts#L4).

The main motivation for doing this is so that plugin-update doesn't have to depend on oclif - which should reduce it's size considerably

### DIFF
--- a/src/interfaces/hooks.ts
+++ b/src/interfaces/hooks.ts
@@ -26,11 +26,11 @@ export interface Hooks {
     return: void;
   };
   preupdate: {
-    options: {channel: string};
+    options: {channel: string, version: string};
     return: void;
   };
   update: {
-    options: {channel: string};
+    options: {channel: string, version: string};
     return: void;
   };
   'command_not_found': {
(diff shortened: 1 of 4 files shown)


src/interfaces/hooks.ts
```diff
@@ -26,11 +26,11 @@ export interface Hooks {
     return: void;
   };
   preupdate: {
-    options: {channel: string};
+    options: {channel: string, version: string};
```
peternhale, line 29: Seems that channel and version are mutually exclusive, so should these be optional?

src/interfaces/hooks.ts
```diff
@@ -26,11 +26,11 @@ export interface Hooks {
...
     return: void;
   };
   update: {
-    options: {channel: string};
+    options: {channel: string, version: string};
```
peternhale, line 33: Ditto

peternhale (changes requested):

src/interfaces/hooks.ts
mdonnalley: Not necessarily because there's a version associated with the channel you're updating too

peternhale (approved):

PR #368, opened 2022-02-09
fix: update isProd utility

### LINKED ISSUE #367: Tests run against lib, not src
Hi,

After we bumped our [CLI](https://github.com/coveo/cli/tree/master/packages/cli) to @oclif/core, the tests are no longer running against the `src/` folder. Instead, they are using the transpiled JavaScript from the lib folder.

This is absolutely fine for testing basic commands. However, mocking modules no longer works if the tested commands are coming from the `lib/` and not the `src/` folder.

When the tests are running, the environment variable `NODE_ENV` is set to `test`. Which will cause the method in [`src/util.ts#isProd`](https://github.com/oclif/core/blob/main/src/util.ts#L39) to always always return `true` when running the tests, and therefore use the `lib/` instead of the `src/` folder.

Since I am not sure this condition was designed like this in purpose, I manually set `NODE_ENV` to `development` just before running the tests.

Should the method `isProd()` also check the the `NODE_ENV` is not equal to `test`?

Thanks,

FYI, we are using Jest instead of mocha.

mdonnalley:
`isProd` returns false if `process.env.NODE_ENV` is either `development` or `test`

Fixes #367

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -37,5 +37,5 @@ export function castArray<T>(input?: T | T[]): T[] {
 }
 
 export function isProd() {
-  return process.env.NODE_ENV !== 'development'
+  return !['development', 'test'].includes(process.env.NODE_ENV ?? '')
 }


y-lakhdar:
Thanks 👍

src/util.ts
```diff
@@ -37,5 +37,5 @@ export function castArray<T>(input?: T | T[]): T[] {
 }
 
 export function isProd() {
-  return process.env.NODE_ENV !== 'development'
+  return !['development', 'test'].includes(process.env.NODE_ENV ?? '')
```
RodEsp, line 40: Can we please try to err on the side of more readable code?

I would much rather see something like `return process.env.NODE_ENV !== 'development && process.env.NODE_ENV !== 'test'`. It's so much easier to parse what exactly that code does and intends to do than something like this.

RodEsp: @mdonnalley, @peternhale fyi ^

peternhale: How about?
`return !['development', 'test'].some(mode => mode === (process.env.NODE_ENV ?? '')`

y-lakhdar: I think we can go even simpler. How about?
`return !['development', 'test'].reduce((prev, curr) => curr === (process.env.NODE_ENV ?? '') || prev, false)`

PR #379, opened 2022-02-24
fix: rid core of transient refs to cli-ux

peternhale:
@W-0@
Bumped deps to remove transient deps on cli-ux

Starting to see the "dependency spiral of death" again with cli-ux


package.json
```diff
@@ -37,49 +37,49 @@
...
+    "shelljs": "^0.8.5",
+    "shx": "^0.3.4",
     "sinon": "^11.1.2",
     "ts-node": "^9.1.1",
-    "typescript": "4.4.4"
+    "typescript": "4.5.5"
   },
   "resolutions": {
-    "@oclif/command": "1.8.9",
+    "@oclif/command": "1.8.16",
```
mdonnalley, line 82: Do we still need these resolutions?

mdonnalley (dismissed):

package.json
peternhale: Removed the resolution for command

mdonnalley (approved):

RodEsp (approved):

PR #384, opened 2022-03-01
feat: dir and file flags that validate existence and type

mshanemc:
dir and file flags from https://github.com/salesforcecli/sf-plugins-core/pull/32

original WI: [@W-10766801@](https://gus.lightning.force.com/a07EE00000roHVgYAM)

### DIFF
--- a/src/parser/flags.ts
+++ b/src/parser/flags.ts
@@ -43,19 +43,19 @@ export const integer = build({
   },
 })
 
-/**
- * Accepts a directory path.  Validates that the directory exists and is a directory.
- */
-export const existingDirectory = build({
-  parse: async (input: string) => dirExists(input),
-})
+export const directory = (opts: { exists?: boolean } & Partial<OptionFlag<string>> = {}): OptionFlag<string | undefined> => {
+  return build<string>({
+    ...opts,
+    parse: async (input: string) => opts.exists ? dirExists(input) : input,
+  })()
+}
 
-/**
- * Accepts a file path.  Validates that the file exists and is a file.
- */
-export const existingFile = build({
-  parse: async (input: string) => fileExists(input),
-})
+export const file = (opts: { exists?: boolean } & Partial<OptionFlag<string>> = {}): OptionFlag<string | undefined> => {
+  return build<string>({
+    ...opts,
+    parse: async (input: string) => opts.exists ? fileExists(input) : input,
+  })()
+}
 
 /**
  * Initializes a string as a URL. Throws an error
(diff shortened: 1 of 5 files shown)


src/parser/flags.ts
```diff
@@ -42,6 +43,20 @@ export const integer = build({
   },
 })
 
+/**
+ * Accepts a directory path.  Validates that the directory exists and is a directory.
+ */
+export const existingDirectory = build<string>({
```
mdonnalley, line 49: What do you think about changing these to `directory` and `file` instead? And then you can specify `exists: true` in the configuration if you need the directory or file to exist

mdonnalley (changes requested):

src/parser/flags.ts
mshanemc: what would they do if they weren't `exists: true` ?

Also, none of the other other flags have a configuration beyond the OptionFlag stuff.  We *could* make them more like what we're doing in the sf-core-plugins where there's a need to pass in extra parameters (wrapping flag.build and returning the result of calling it) but that's not how the other oclif flags formed from OptionFlag work.

RodEsp: If `exists: false` it would just validate that you're giving it a valid dir/file path.

mdonnalley (approved):

src/parser/flags.ts
```diff
@@ -42,6 +43,20 @@ export const integer = build({
   },
 })
 
+export const directory = (opts: { exists?: boolean } & Partial<OptionFlag<string>> = {}): OptionFlag<string | undefined> => {
+  return build<string>({
+    ...opts,
+    parse: async (input: string) => opts.exists ? dirExists(input) : input,
```
RodEsp, line 49: If `exists` is `false` this should at least validate the input is a valid directory path.

src/parser/flags.ts
```diff
@@ -42,6 +43,20 @@ export const integer = build({
...
+  })()
+}
+
+export const file = (opts: { exists?: boolean } & Partial<OptionFlag<string>> = {}): OptionFlag<string | undefined> => {
+  return build<string>({
+    ...opts,
+    parse: async (input: string) => opts.exists ? fileExists(input) : input,
```
RodEsp, line 56: If `exists` is `false` this should at least validate the input is a valid file path.

RodEsp (changes requested):

src/parser/flags.ts
mshanemc: can you explain what these validation would do (ex: what's valid vs. invalid?). Like certain characters that can't be used depending on the os?

RodEsp: @mshanemc yep, exactly that.

RodEsp (approved):

mdonnalley (approved):

PR #386, opened 2022-03-02
Sm/help-default-fix

mshanemc:
adds a property to Flags so that if
1) using default that's a function
2) flag type is not string
3) flag can return the result of another executed function that is a string

with support for fallbacks and faillures so that toCached (from manifest build process) won't throw

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -7,7 +7,7 @@ import {format} from 'util'
 
 import {Options, Plugin as IPlugin} from '../interfaces/plugin'
 import {Config as IConfig, ArchTypes, PlatformTypes, LoadOptions} from '../interfaces/config'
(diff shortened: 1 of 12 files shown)


src/config/config.ts
```diff
@@ -560,6 +560,27 @@ export class Config implements IConfig {
   }
 }
 
+// when no manifest exists, the default is calculated.  This may throw, so we need to catch it
+const defaultToCached = async (flag: CompletableOptionFlag<any>) => {
+  // Prefer the helpDefaultValue function (returns a friendly string for complex types)
+  if (typeof flag.defaultHelp === 'function') {
```
mdonnalley, line 566: Isn't the goal of this property to be calculated at run time? If so, it needs to be rerun whenever help is generated: https://github.com/oclif/core/blob/main/src/help/command.ts#L290

peternhale (approved):

PR #404, opened 2022-04-19
[POC] feat: show aliases under their own help section

RodEsp:
## What does this PR do?
This makes aliases show up under their own section in the help output for a CLI instead of having them show up under the COMMANDS.

image

## Testing instructions
1. Get this branch locally
    1. Run `yarn install`
    1. Run `yarn link`
1. Clone `https://github.com/RodEsp/test-cli` locally
    1. Run `yarn install`
    1. Run `yarn link @oclif/core`
    1. Run `bin/dev help`
1. See `hello-alias` show up under `ALIASES` instead of `COMMANDS`


purplecabbage:
I sort of solved (part of) the same issue here: #492 
In our case we have many aliases, so we only want them to show up when you do --help on the full subCommand.

If we used your solution we would get a crazy list as we have multilevel aliases.
for example:
```
aio runtime namespace list
aio runtime namespace ls
aio runtime ns list
aio runtime ns ls
aio rt namespace list
aio rt namespace ls
aio rt ns list
aio rt ns ls
```

PR #426, opened 2022-06-13
fix: support CLIs with single top level command

### LINKED ISSUE #425: Migrate single command cli to @oclif/core
Hi guys, I'm trying to migrate [bindl](https://github.com/felipecrs/bindl), but I'm having a really hard time.

It's a single command CLI (i.e. just calling `bindl` would do the trick), but after following all the advices in the migration guide, it simply does not work.

Am I missing something?

```console
# Before
❯ bin/run
 ›   Error: I was not able to load the configuration file.

# After
❯ bin/dev
Downloads and extracts binaries from compressed packages using a config file

VERSION
  bindl/2.0.0 wsl-x64 node-v16.15.1

USAGE
  $ bindl [COMMAND]
```

mdonnalley:
Support CLIs with a single top level command (i.e. the executable is the only command).

With this fix, users will need to do two things:
1. Place their command at the top level of the `commands` dir, e.g. `src/commands/index.ts`
2. supply the `default` in their oclif config, for example

```
oclif: {
  default: ".",
  commands: "./dist/commands"
}
```

Fixes #425

### DIFF
--- a/src/config/plugin.ts
+++ b/src/config/plugin.ts
@@ -192,7 +192,8 @@ export class Plugin implements IPlugin {
       const p = path.parse(file)
       const topics = p.dir.split('/')
       const command = p.name !== 'index' && p.name
-      return [...topics, command].filter(f => f).join(':')
+      const id = [...topics, command].filter(f => f).join(':')
+      return id === '' ? '.' : id
     })
     this._debug('found commands', ids)
     return ids


felipecrs:
Please let me know if the below migration would be correct then:

1. Add to `package.json`:

```json
{
  "oclif": {
    "default": ".",
    "commands": "./lib" // as my "index.ts" is at "src/index.ts" and my tsc builds to "lib"
  }
}
```

mdonnalley:
> Please let me know if the below migration would be correct then:
> 
> 1. Add to `package.json`:
> 
> ```json
> {
>   "oclif": {
>     "default": ".",
>     "commands": "./lib" // as my "index.ts" is at "src/index.ts" and my tsc builds to "lib"
>   }
> }
> ```

That should also work

felipecrs:
It works, but the `--help` now displays two sections (while it should only display the first):

```console
❯ bin/dev --help
Downloads and extracts binaries from compressed packages using a config file

USAGE
  $ bindl . [--version] [--help] [-c <value>]

FLAGS
  -c, --config=<value>  Path to the config file
  --help                Show CLI help.
  --version             Show CLI version.

DESCRIPTION
  Downloads and extracts binaries from compressed packages using a config file

  The config will be read from any valid config file in the current directory. The configuration file can be defined
  using all the extensions and names accepted by cosmiconfig, such as bindl.config.js



Downloads and extracts binaries from compressed packages using a config file

VERSION
  bindl/2.0.0 wsl-x64 node-v16.15.1

USAGE
  $ bindl [COMMAND]

COMMANDS
  .  Downloads and extracts binaries from compressed packages using a config file

```

felipecrs:
Another issue:

`bindl .` triggers the command too (while it should actually pass `.` as argument to the main command.

mdonnalley:
@felipecrs I push a commit last night that should resolve the issues you're seeing - let me know if that works better for you

RodEsp (approved):

PR #441, opened 2022-07-14
feat: add stderr method

### LINKED ISSUE #392: Support logging to stderr
Command has a `log()` method for logging to stdout. Part of being a well-behaved CLI program is only logging stuff to stdout that might be consumed by another process, e.g. fed into another program's stdin. That means errors and other debug/status info should be printed to stderr. It would be nice if oclif commands supported this like they do with stdout via `this.log()`

As it stands, I have to use `process.stderr` (which is what Command.log is doing under the hood for stdout)

mdonnalley:
Adds `stderr` method to `Command` for logging to stderr

Fixes #392

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -199,22 +199,18 @@ export default abstract class Command {
     return Errors.error(input, options as any)
   }
 
-  stdout(message = '', ...args: any[]): void {
+  log(message = '', ...args: any[]): void {
     if (!this.jsonEnabled()) {
       message = typeof message === 'string' ? message : inspect(message)
       process.stdout.write(format(message, ...args) + '\n')
     }
   }
 
-  stderr(message = '', ...args: any[]): void {
+  logToStderr(message = '', ...args: any[]): void {
     if (!this.jsonEnabled()) {
       message = typeof message === 'string' ? message : inspect(message)
       process.stderr.write(format(message, ...args) + '\n')
     }
-  }
-
-  log(message = '', ...args: any[]): void {
-    this.stdout(message, ...args)
   }
 
   public jsonEnabled(): boolean {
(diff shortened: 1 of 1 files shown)


src/command.ts
```diff
@@ -199,14 +199,24 @@ export default abstract class Command {
     return Errors.error(input, options as any)
   }
 
-  log(message = '', ...args: any[]): void {
+  stdout(message = '', ...args: any[]): void {
```
RodEsp, line 202: I think we should just keep this as `log`. I don't want to confuse anyone by using the name `stdout` since that's already a well known method in a node process and this method doesn't behave exactly like it.

src/command.ts
```diff
@@ -199,14 +199,24 @@ export default abstract class Command {
...
     if (!this.jsonEnabled()) {
-      // tslint:disable-next-line strict-type-predicates
       message = typeof message === 'string' ? message : inspect(message)
       process.stdout.write(format(message, ...args) + '\n')
     }
   }
 
+  stderr(message = '', ...args: any[]): void {
```
RodEsp, line 209: Similarly to my comment above I'd rather not name this `stderr`. Perhaps we can go with use `error()` or `logError()`?

RodEsp (changes requested):

src/command.ts
RodEsp: I don't think using `logError()` and the implicit assumption of only logging errors to `stderr` is necessarily a bad thing. It's the same assumption the POSIX standard makes by naming that stream `stderr` instead of `nonstdout` or something that encompasses more than just errors.

If users want to use `stderr` for something other than errors that's up to them.

That being said, I'm fine with using `logToStderr` if you'd rather do that since my whole point here is just to make it clear this method is something different than `process.stderr` and `logToStderr` achieves that.

RodEsp (approved):

PR #447, opened 2022-07-20
feat: min/max for integer flag

mshanemc:


src/parser/flags.ts
```diff
@@ -35,13 +35,21 @@ export function boolean<T = boolean>(
...
-  },
-})
+export const integer = (opts: { min?: number; max?: number } & Partial<OptionFlag<number>> = {}): OptionFlag<number | undefined> => {
+  return build<number>({
+    ...opts,
+    parse: async input => {
+      if (!/^-?\d+$/.test(input))
+        throw new Error(`Expected an integer but received: ${input}`)
+      const num = Number.parseInt(input, 10)
+      if (typeof opts.min === 'number' && Number(input) < opts.min)
```
peternhale, line 45: Why parse the input again?

peternhale (changes requested):

src/parser/flags.ts
```diff
@@ -35,13 +35,21 @@ export function boolean<T = boolean>(
...
 }
 
-export const integer = build({
-  parse: async input => {
-    if (!/^-?\d+$/.test(input))
-      throw new Error(`Expected an integer but received: ${input}`)
-    return Number.parseInt(input, 10)
-  },
-})
+export const integer = (opts: { min?: number; max?: number } & Partial<OptionFlag<number>> = {}): OptionFlag<number | undefined> => {
```
peternhale, line 38: It would be nice if min and max were also integers, maybe change type to `bigint`?

mdonnalley (dismissed):

peternhale (dismissed):

src/parser/flags.ts
```diff
@@ -35,13 +35,21 @@ export function boolean<T = boolean>(
...
+    ...opts,
+    parse: async input => {
+      if (!/^-?\d+$/.test(input))
+        throw new Error(`Expected an integer but received: ${input}`)
+      const inputInt = BigInt(input)
+      if (typeof opts.min === 'bigint' && inputInt < opts.min)
+        throw new Error(`Expected an integer greater than or equal to ${opts.min} but received: ${input}`)
+      if (typeof opts.max === 'bigint' && inputInt > opts.max)
+        throw new Error(`Expected an integer less than or equal to ${opts.max} but received: ${input}`)
+      return Number(inputInt)
```
RodEsp, line 49: Why make these BigInts instead of normal numbers?
I'd recommend against it unless there's a good reason, and if so we should put it in a comment. Also, if they have to be BigInts we shouldn't cast them back to a Number in the return statement because they can lose precision that way.

There's a few pitfalls associated with BigInts, [see this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt).

RodEsp (changes requested):

src/parser/flags.ts
mshanemc: "Pete said to" is really all I got.  I'd prefer my original version with numbers.  Will switch it back now.

They needed to end up as a number to avoid breaking changes to the oclif integer flag, which results in type `number`

RodEsp: Sounds good. If need be, we can just create a `BigInt` flag instead of changing this one.

peternhale (dismissed):

src/parser/flags.ts
```diff
@@ -35,13 +35,21 @@ export function boolean<T = boolean>(
...
 }
 
-export const integer = build({
-  parse: async input => {
-    if (!/^-?\d+$/.test(input))
-      throw new Error(`Expected an integer but received: ${input}`)
-    return Number.parseInt(input, 10)
-  },
-})
+export const integer = (opts: { min?: bigint; max?: bigint } & Partial<OptionFlag<number>> = {}): OptionFlag<number | undefined> => {
```
RodEsp, line 38:
```suggestion
export const integer = (opts: { min?: int; max?: int} & Partial<OptionFlag<number>> = {}): OptionFlag<number | undefined> => {
```

src/parser/flags.ts
```diff
@@ -35,13 +35,21 @@ export function boolean<T = boolean>(
...
+    ...opts,
+    parse: async input => {
+      if (!/^-?\d+$/.test(input))
+        throw new Error(`Expected an integer but received: ${input}`)
+      const num = Number.parseInt(input, 10)
+      if (typeof opts.min === 'number' && num < opts.min)
+        throw new Error(`Expected an integer greater than or equal to ${opts.min} but received: ${input}`)
+      if (typeof opts.max === 'number' && num > opts.max)
+        throw new Error(`Expected an integer less than or equal to ${opts.max} but received: ${input}`)
+      return num
```
RodEsp, line 49: If I'm reading this correctly if `opts.min` or `opts.max` aren't typeof number they'll just be ignored and we'll return w/e the user passed in?

Should probs add an error if they're something other than typeof number (yeah I know TS should avoid this but what if someone doesn't have TS setup on their editor?).

RodEsp (changes requested):

src/parser/flags.ts
mshanemc: int doesn't exist.  But I'm gonna make them `number`.  good catch.

src/parser/flags.ts
mshanemc: decision from synchronous discussion:

typeof was handling the optional properties, but excluding `undefined` would allow runtime errors on min/max for non-number types that Rod wants.

src/parser/flags.ts
RodEsp: Whoops! 👍🏼

RodEsp (approved):

src/parser/flags.ts
RodEsp: Just to clarify, it's not that I _want_ this to error but the developer should be made aware if they are using non-numerical values for `min`/`max` that they can't do that. That wouldn't have happened with the previous `typeof min === 'number'` check. It would have just returned w/e the user passed in without comparing it to a min, same for max.

Would be ideal if these could be compile time errors and that's why we use TS but even that won't catch all scenarios.

jasonmobley:
The tsconfig.json change in this issue was a breaking change for anyone still using Node v12, but the subsequent release was only a minor version bump (does this project not follow semver?). The use of optional chaining operator is no longer transpiled out by tsc, and Node 12 doesn't support that operator. Removing a supported version from engines should be a breaking change!

PR #448, opened 2022-07-21
fix: flush not hitting drain condition

modi98:
This fixes the flush function never resolving, `process.stdout.write('')` will return true if the writing chunk was accepted, only if it returns false we'll need to listen to the drain event.

More info on the Stream documentation -> https://nodejs.org/docs/latest-v14.x/api/stream.html#stream_writable_write_chunk_encoding_callback


modi98:
Signed in to salesforce-cla 👍

mdonnalley (approved):

PR #450, opened 2022-07-21
fix: flag setter order

### LINKED ISSUE #449: order of enableJsonFlag in code affects whether or not flags are registered
**Describe the bug**
Specifying enableJsonFlag after flags, causes the flags to not register. This worked fine in the last version I tried, 1.9.5, but has changed in 1.12.0.

**To Reproduce**

will only register the json flag
```javascript
static flags = {
  test: Flags.string()
}
static enableJsonFlag = true
```

will register both flags
```javascript
static enableJsonFlag = true
static flags = {
  test: Flags.string()
}
```

mdonnalley:
Make order of `enableJsonFlag` and `flags` declaration irrelevant

Fixes #449

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -136,7 +136,7 @@ export default abstract class Command {
 
   static set globalFlags(flags: Interfaces.FlagInput<any>) {
     this._globalFlags = Object.assign({}, this.globalFlags, flags)
-    this.flags = this.globalFlags
+    this.flags = {} // force the flags setter to run
   }
 
   /** A hash of flags for the command */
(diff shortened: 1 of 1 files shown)


src/command.ts
```diff
@@ -136,7 +136,7 @@ export default abstract class Command {
 
   static set globalFlags(flags: Interfaces.FlagInput<any>) {
     this._globalFlags = Object.assign({}, this.globalFlags, flags)
-    this.flags = this.globalFlags
+    this.flags = {} // force the flags setter to run
```
peternhale, line 139: @nsamsami Not sure I understand the "force the setter to run" as before and after  changes  are both simple assignments?

mdonnalley: In both cases, the setter was being executed. But I changed it to an empty object since the flags setter is already adding `this.globalFlags`. In other words, it's duplicative to pass in `this.globalFlags`

I added the comment in case anyone looked at this and thought we were just setting `this.flags` to an empty object

peternhale (approved):

src/command.ts
RodEsp: 🤔 
Yeah, this comment seems kinda weird. Could you elaborate @mdonnalley?

PR #457, opened 2022-08-05
fix: handle enableJsonFlag value of false

peternhale:


src/command.ts
```diff
@@ -104,7 +104,13 @@ export default abstract class Command {
 
   static set enableJsonFlag(value: boolean) {
     this._enableJsonFlag = value
-    if (value) this.globalFlags = jsonFlag
+    if (value) {
```
peternhale, line 107: @mdonnalley For a subclass where base class enables json and then the subclass disables json, second time through (value === false) needs to be applied global.

mdonnalley: I think the following might be a cleaner solution.

1. Remove the getter/setter for `enableJsonFlag`. So now you just have a normal static class property

```
static enableJsonFlag = false
```

2. Move the json flag addition to `Command.parse` by replacing the assignment of `opts.flags` with the following
```
opts.flags = (this.ctor.enableJsonFlag ? {...options?.flags, ...jsonFlag} : options?.flags) as Interfaces.FlagInput<F>
```

mdonnalley (changes requested):

src/command.ts
```diff
@@ -131,7 +122,7 @@ export default abstract class Command {
   protected static _globalFlags: Interfaces.FlagInput
 
   static get globalFlags(): Interfaces.FlagInput {
-    return this._globalFlags
+    return Object.assign(this._globalFlags ?? {}, this.enableJsonFlag ? {...jsonFlag} : {})
```
mdonnalley, line 125: If you merge the json flag here, the order in which `enableJsonFlag` and `globalFlags` are set will matter - which we don't want, see #449

mdonnalley (changes requested):

src/command.ts
peternhale, line 125: I placed the eval of enableJsonFlag here so tests would pass. The format tests were failing missing the Global Flags section.

PR #460, opened 2022-08-08
fix: retain enable json flag get/set apply flag at parse

peternhale:


mdonnalley:
QA

`Command` class with `enableJsonFlag = true`
✅ run command with --help, shows `--json`
✅ run command with --json, succeeds
✅ order of `enableJsonFlag` and `flags` does not matter

Subclass extends Parent `Command` class with `enableJsonFlag = true`
✅ run command with --help, shows `--json`
✅ run command with --json, succeeds
✅ order of `enableJsonFlag` and `flags` does not matter

Subclass with `enableJsonFlag = false` extends Parent `Command` class with `enableJsonFlag = true`
✅ run command with --help, does not show `--json`
✅ run command with --json, fails
✅ order of `enableJsonFlag` and `flags` does not matter

mdonnalley (approved):

PR #468, opened 2022-08-10
feat: support complex flag relationships

mdonnalley:
Lay groundwork to support more advanced flag relationship validation

Example usage

```typescript

  static flags = {
    one: Flags.boolean(),
    two: Flags.boolean(),
    three: Flags.boolean(),
    foo: Flags.boolean({
      relationships: {
        dependsOn: {
          type: 'atLeastOne',
          flags: ['one', 'two'],
        },
        exclusive: { type: 'all', flags: ['three'] },
      },
    }),
  }
}
```

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -751,6 +751,7 @@ export async function toCached(c: Command.Class, plugin?: IPlugin): Promise<Comm
         helpGroup: flag.helpGroup,
         allowNo: flag.allowNo,
         dependsOn: flag.dependsOn,
+        relationships: flag.relationships,
         exclusive: flag.exclusive,
       }
     } else {
@@ -768,6 +769,7 @@ export async function toCached(c: Command.Class, plugin?: IPlugin): Promise<Comm
         multiple: flag.multiple,
         options: flag.options,
         dependsOn: flag.dependsOn,
+        relationships: flag.relationships,
         exclusive: flag.exclusive,
         default: await defaultToCached(flag),
       }
--- a/src/interfaces/parser.ts
+++ b/src/interfaces/parser.ts
@@ -112,6 +112,16 @@ export type FlagProps = {
   hidden?: boolean;
   required?: boolean;
   dependsOn?: string[];
+  relationships?: {
+    dependsOn?: {
+      type: 'all' | 'atLeastOne';
+      flags: string[];
+    };
+    exclusive?: {
+      type: 'all' | 'atLeastOne';
+      flags: string[];
+    }
+  };
   exclusive?: string[];
 }
 
--- a/src/parser/validate.ts
+++ b/src/parser/validate.ts
@@ -6,7 +6,7 @@ import {
   RequiredFlagError,
   UnexpectedArgsError,
 } from './errors'
-import {ParserArg, ParserInput, ParserOutput, Flag} from '../interfaces'
+import {ParserArg, ParserInput, ParserOutput, Flag, CompletableFlag} from '../interfaces'
 
 export function validate(parse: {
   input: ParserInput;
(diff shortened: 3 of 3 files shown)


src/interfaces/parser.ts
```diff
@@ -112,6 +112,16 @@ export type FlagProps = {
   hidden?: boolean;
   required?: boolean;
   dependsOn?: string[];
+  relationships?: {
+    dependsOn?: {
+      type: 'all' | 'atLeastOne';
```
mshanemc, line 117: `none` could allow exclusive to go away?

src/interfaces/parser.ts
```diff
@@ -112,6 +112,16 @@ export type FlagProps = {
...
+      flags: string[];
```
mshanemc, line 118: `flags: string[] | Array<{flag: string, value: any}>` would allow you to say things like:

flag A can't be used if flag B has value X (exclusive)
or
you can't use flag A unless flag B has value Y (dependsOn)

or even (allow or disallow multiple values from a flag)
```
[{ flag: 'a', value: 'x'}, {flag: 'a', value: 'y'}]
```

src/parser/validate.ts
```diff
@@ -57,44 +57,104 @@ export function validate(parse: {
...
+        throw new RequiredFlagError({parse, flag})
+      } else if (flag.exactlyOne && flag.exactlyOne.length > 0) {
+        validateAcrossFlags(flag)
+      }
+    }
+  }
+
+  function validateExclusive(name: string, exclusive: string[]) {
+    for (const also of exclusive) {
+      // do not enforce exclusivity for flags that were defaulted
```
mshanemc, line 74: consider making this configurable.  Flag B exists, but can't be used unless you make Flag A something other than its default

src/parser/validate.ts
```diff
@@ -57,44 +57,104 @@ export function validate(parse: {
...
+      if (flag.relationships?.dependsOn.type === 'all') {
+        validateDependsOn(name, dependsOnFlags)
+      }
+
+      if (flag.relationships.dependsOn.type === 'atLeastOne') {
+        let foundAtLeastOne = false
+        for (const flag of dependsOnFlags) {
+          if (parse.output.flags[flag]) {
+            foundAtLeastOne = true
+            break
```
mshanemc, line 127:
```suggestion
        const foundAtLeastOne = dependsOnFlags.some((flag) =>  parse.output.flags[flag])
```

src/interfaces/parser.ts
```diff
@@ -112,6 +112,16 @@ export type FlagProps = {
...
+    };
+    exclusive?: {
+      type: 'all' | 'atLeastOne';
```
mshanemc, line 121: checking my understanding:

'all' => it's an error if you use all the flags listed?
'atLeastOne' => it's an error if you use any of the flags listed?

src/config/config.ts
```diff
@@ -751,6 +751,7 @@ export async function toCached(c: Command.Class, plugin?: IPlugin): Promise<Comm
         helpGroup: flag.helpGroup,
         allowNo: flag.allowNo,
         dependsOn: flag.dependsOn,
+        relationships: flag.relationships,
```
mshanemc, line 754: would we want to use union types to make it not possible to use both relationships and any of the simple ones (dependsOn, exclusive, exactlyOne) so that it doesn't get really nasty?

mdonnalley: I feel comfortable not putting any guard rails up for that. People can make a mess of their flags if they want

For the sf side, we could always add some linting rules to guide people away from doing too much mixing and matching

src/interfaces/parser.ts
```diff
@@ -86,6 +86,12 @@ export type DefaultContext<T> = {
 export type Default<T> = T | ((context: DefaultContext<T>) => Promise<T>)
 export type DefaultHelp<T> = T | ((context: DefaultContext<T>) => Promise<string | undefined>)
 
+export type FlagRelationship = string | {name: string; when: (flags: Record<string, unknown>) => Promise<boolean>};
+export type Relationship = {
+  type: 'all' | 'some' | 'none';
```
mshanemc, line 91: consider `one` or `1` since we've already got an `exactlyOne` since it happened often enough.

src/parser/validate.ts
```diff
@@ -99,6 +71,96 @@ export function validate(parse: {
...
+    const foundAtLeastOne = Object.values(resolved).some(Boolean)
+    if (!foundAtLeastOne) {
+      const formattedFlags = Object.keys(resolved).map(f => `--${f}`).join(', ')
+      throw new CLIError(`One of the following must be provided when using --${name}: ${formattedFlags}`)
+    }
+  }
+
+  async function validateRelationships(name: string, flag: CompletableFlag<any>) {
+    if (!flag.relationships) return
+    for (const relationship of flag.relationships) {
```
mshanemc, line 147: worth parallelizing?

src/parser/validate.ts
```diff
@@ -99,6 +71,96 @@ export function validate(parse: {
...
+      if (parse.output.flags[flag]) {
+        throw new CLIError(
+          `--${flag}=${parse.output.flags[flag]} cannot also be provided when using --${name}`,
+        )
+      }
+    }
+  }
+
+  function validateExactlyOne(name: string, exactlyOne: FlagRelationship[]) {
+    for (const flag of exactlyOne || []) {
```
mshanemc, line 115: defaults in the signature
```suggestion
  function validateExactlyOne(name: string, exactlyOne: FlagRelationship[] = []) {
    for (const flag of exactlyOne) {
```

src/parser/validate.ts
mshanemc: ux idea: as written, you'll get a failure on the first validation to throw.  For the user, they'll fix one error and then hit the next.

You could run this with Promise.allSettled and maybe display all the errors in the output.

src/parser/validate.ts
```diff
@@ -54,43 +56,13 @@ export function validate(parse: {
     }
   }
 
-  function validateFlags() {
+  async function validateFlags() {
     for (const [name, flag] of Object.entries(parse.input.flags)) {
```
mshanemc, line 60: I'd love to see if parallelizing this would help with perf (both the various validations for each flag, AND doing all the flags together)

src/parser/validate.ts
```diff
@@ -99,6 +71,96 @@ export function validate(parse: {
...
+    const resolved = await Promise.all(promises)
+    return Object.fromEntries(resolved.filter(r => r !== null) as [string, unknown][])
+  }
+
+  async function validateExclusive(name: string, flags: FlagRelationship[]) {
+    const resolved = await resolveFlags(flags)
+    const keys = Object.keys(resolved).reduce((acc, key) => {
+      if (resolved[key]) acc.push(key)
+      return acc
+    }, [] as string[])
```
mshanemc, line 92: not sure why this reduce needs to be here.  Another design would be

Object.entries(resolved).forEach(

and use 1 && conditional to check what you're checking to get to an error
1. the value is truthy AND
2.`parse.output.metadata.flags[flag]?.setFromDefault` is falsy AND
3. `parse.output.metadata.flags[name]?.setFromDefault` is falsy AND
4. `parse.output.flags[flag]` is truthy
--> throw!

that way it's only 1 iteration, the conditionals can be in whatever order is most likely to resolve soonest and it'll fail on the first error

mshanemc (dismissed):
approved with some ideas/suggestions

PR #471, opened 2022-08-16
fix: more custom flag type overloads

### LINKED ISSUE #469: Bug: type inference for flags with `multiple: true`
**Describe the bug**

See [this TS playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgYQiEBDAdgEwDRwBiANhgOYDOcAvnAGZRpwDkAAhAMbHB0D0H0AKbMAUCMEAPSLDg5BdDAFdi8LhgpVCECHEkxBuKqnTYciEXDgUYGGMA6zBFDlGBg7ELHAC8LACoAnmCCcMBYdIJQBhwhwFQARowA1gb00JZ0pJRwAO7AMAAWcAAGIMp2YMSCxQB0zADcYpbWtvb0WVS+CBaWcFiKIPGRFABcRB01YfpkkQAU3b29HAUYUGPMWMx4PYtyzq7uwJ7rAIJw3NZwEHR9A0NQFFs7vWUqblVjMFCKgtuL1ABKP69ayuLCUMYkcgUGqgsJkebPSzLVbrR7A3ZOFxuDxYU7nOLwa5WL7w9FIuCvCofOBfH4YmhAnrURo9MCKeLcBzqAJYBzfLCzAFjAAKjBAcUEAB4AG4QYA4AB85kWAiwlyQmWhNB8cAwOQw+VpBTiNTAqwoglmWggANZi14vFpQRCxK1lBq-UGw1CVGKXvucAAPnBFLh5GFBDhis9Hc7glcbu6YXDwVQ4iVU2QANoAXRjsad8QghT1UVCfOgUQ4MGZImoQA) to inspect the type inference.

```ts
import { Command, Flags } from '@oclif/core'

export default class Foo extends Command {
  static description = 'Type inference is broken for flags with `multiple`.';

  static flags = {
    numbers: Flags.integer({
      char: 'n',
      description: 'A list of numbers',
      multiple: true,
    }),
    strings: Flags.string({
      char: 's',
      description: 'A list of strings',
      multiple: true,
    }),
  };

  public async run(): Promise<void> {
    const { flags } = await this.parse(Foo);

    // type of flags.numbers is `number | undefined`
    // type of flags.strings is `string[]`

    // both are incorrect
  }
}
```

**Expected behavior**

With `multiple: true`, it should infer the flag's type as `T[] | undefined`. For `Flags.integer` it doesn't infer the array correctly, and for `Flags.string` it's missing the `| undefined` union.

(The real type of the flags can easily be inspected with `console.log(flags)`.)

BTW the inference was working correctly after we bootstrapped our project, because it was using `@oclif/core@1.0.10` but then broke after upgrading it to the latest version, `@oclif/core@1.13.10`. I'm using macOS (don't think it matters) and TypeScript version `4.7.4`.

**Environment**
 - OS & version: MacOS Monterey

**Additional context**

Downgraded to latest `1.11` and there it's working except it's missing the `| undefined` union as well.

mdonnalley:
Fixes #469

### DIFF
--- a/src/interfaces/parser.ts
+++ b/src/interfaces/parser.ts
@@ -149,7 +149,11 @@ export type OptionFlag<T> = FlagBase<T, string> & OptionFlagProps & {
 }
 
 export type Definition<T> = {
-  (options: { multiple: true } & Partial<OptionFlag<T[]>>): OptionFlag<T[]>;
+  (
+    options: { multiple: true } & ({ required: true } | { default: Default<T> }) &
+      Partial<OptionFlag<T>>,
+  ): OptionFlag<T[]>;
+  (options: { multiple: true } & Partial<OptionFlag<T[]>>): OptionFlag<T[] | undefined>;
   (
     options: ({ required: true } | { default: Default<T> }) &
       Partial<OptionFlag<T>>,
--- a/src/parser/flags.ts
+++ b/src/parser/flags.ts
@@ -35,9 +35,11 @@ export function boolean<T = boolean>(
   } as BooleanFlag<T>
 }
 
+export function integer(opts: Partial<OptionFlag<number>> & {min?: number; max?: number } & {multiple: true} & ({required: true} | { default: Default<number> })): OptionFlag<number[]>
+export function integer(opts: Partial<OptionFlag<number>> & {min?: number; max?: number } & {multiple: true}): OptionFlag<number[] | undefined>
 export function integer(opts: Partial<OptionFlag<number>> & {min?: number; max?: number } & ({required: true} | { default: Default<number> })): OptionFlag<number>
 export function integer(opts?: Partial<OptionFlag<number>> & {min?: number; max?: number }): OptionFlag<number | undefined>
-export function integer(opts: Partial<OptionFlag<number>> & {min?: number; max?: number } = {}): OptionFlag<number> | OptionFlag<number | undefined> {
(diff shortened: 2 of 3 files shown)


src/interfaces/parser.ts
```diff
@@ -149,7 +149,11 @@ export type OptionFlag<T> = FlagBase<T, string> & OptionFlagProps & {
 }
 
 export type Definition<T> = {
-  (options: { multiple: true } & Partial<OptionFlag<T[]>>): OptionFlag<T[]>;
+  (
+    options: { multiple: true } & ({ required: true } | { default: Default<T> }) &
+      Partial<OptionFlag<T>>,
+  ): OptionFlag<T[]>;
+  (options: { multiple: true } & Partial<OptionFlag<T[]>>): OptionFlag<T[] | undefined>;
```
allan-barbato, line 156: @mdonnalley I think this change broke my code.

I have a command setup like:
```
class MyCommand extends Command {
  static flags: FlagInput<{ include: string[] }> = {
    include: Flags.string({
      multiple: true,
      default: [],
    }),
  }
}
```
And now it's giving me an error like: `Type 'OptionFlag<string[] | undefined>' is not assignable to type 'CompletableFlag<string[]>'`

I works if I edit the `parser.d.ts` file manually like so:
```
  (
    options: { multiple: true } & ({ required: true } | { default: Default<T[]> }) &
      Partial<OptionFlag<T[]>>,
  ): OptionFlag<T[]>;
```

PR #486, opened 2022-09-01
fix: isProd checks if manifest exists instead of `NODE_ENV` env var.

cristiand391:
Restores old oclif/config behaviour:
https://github.com/oclif/config/blob/main/src/config.ts#L280

### before
oclif would try to load devPlugins if `NODE_ENV=development|test`, this seems correct but prod CLIs don't ship devDeps so trying to load devPlugins makes oclif emit warns like this:
```
NODE_ENV=development sfdx
(node:15374) Error Plugin: sfdx-cli: could not find package.json with {
  type: 'dev',
  root: '/usr/local/lib/sfdx',
  name: '@oclif/plugin-command-snapshot'
}
module: @oclif/core@1.13.10
task: loadPlugins
plugin: sfdx-cli
root: /usr/local/lib/sfdx
See more details with DEBUG=*
(Use `node --trace-warnings ...` to show where the warning was created)
(node:15374) Error Plugin: sfdx-cli: could not find package.json with {
  type: 'dev',
  root: '/usr/local/lib/sfdx',
  name: '@salesforce/plugin-release-management'
}
```

### after
oclif will load devPlugins if a manifest is present at the root CLI dir.

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -24,6 +24,15 @@ function channelFromVersion(version: string) {
   return (m && m[1]) || 'stable'
 }
 
+function hasManifest(p: string): boolean {
+  try {
+    require(p)
+    return true
+  } catch {
+    return false
+  }
+}
+
 const WSL = require('is-wsl')
 
 function isConfig(o: any): o is IConfig {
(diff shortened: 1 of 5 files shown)


src/config/config.ts
```diff
@@ -221,7 +230,7 @@ export class Config implements IConfig {
   async loadDevPlugins() {
     if (this.options.devPlugins !== false) {
       // do not load oclif.devPlugins in production
-      if (this.isProd) return
```
cristiand391, line 224: I wonder if `isProd` should check the existence of a manifest (and/or other files) instead of just `NODE_ENV`.

not sure where that method would live (or can be a func that accepts `this.root`).

cristiand391: first versions of core were checking for a manifest too, changed here: https://github.com/oclif/core/pull/129/files#diff-615dacd7000fa06d7afef0f22b7b6d9267c6219fbfc9fa0315e51a5f5d963755L186

Issue #503, opened 2022-09-29
Deno Support for Packaging


AndresPrez:
[Deno](https://deno.land/manual@v1.26.0/introduction) is a **A modern runtime for JavaScript and TypeScript**, which is becoming increasingly popular. 

The problem is that it has [limitations at the time of compiling into an executable](https://deno.land/manual@v1.26.0/tools/compiler#unavailable-in-executables)

----
We would like to leverage oclif's package command to allow packaging of Deno projects.

XantreGodlike:
What are the progress on that? When can we run oclif with deno?)

AndresPrez:
Hopefully they're working on it. I made a PR for it but I couldn't get passed this signing point: https://github.com/oclif/core/pull/504#issuecomment-1262365040

mdonnalley:
Deno support is not currently on our roadmap because there's simply not enough demand for it and we (that is, Salesforce) don't have any plans on using deno for our own CLI.

Happy to reconsider it though if more people start asking for this.

aelesbao:
I'm definitely interested in running `oclif` using a faster execution engine :)

calebAtIspot:
> because there's simply not enough demand for it 

Note this is currently the second most voted feature by 👍 's

PR #504, opened 2022-09-29
Paves the way for deno oclif packaging

AndresPrez:
Tackles the `oclif/core` https://github.com/oclif/core/issues/503.

It paves the way to start supporting deno projects. It essentially extends the logic around using the `package.json` as the project's root config file to also support Deno's `deno.jsonc` configuration file.

### DIFF
--- a/src/config/util.ts
+++ b/src/config/util.ts
@@ -1,4 +1,6 @@
 import * as fs from 'fs'
+import * as path from 'path'
+import * as JSON5 from 'json5'
 
 const debug = require('debug')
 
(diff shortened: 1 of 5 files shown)


src/config/util.ts
```diff
@@ -23,13 +25,16 @@ export function resolvePackage(id: string, paths: { paths: string[] }): string {
   return require.resolve(id, paths)
 }
 
-export function loadJSON(path: string): Promise<any> {
+export function loadJSON(_path: string): Promise<any> {
   debug('config')('loadJSON %s', path)
+  // Allows reading JSON with comments
+  const _JSON = path.extname(_path) === '.jsonc' ? JSON5 : JSON
```
rektide, line 31: thinking out loud here, maybe using dynamic imports here (so json5 isn't loaded by json users) might be kind.

PR #511, opened 2022-09-30
feat: support flag and command deprecations

mdonnalley:
Support deprecation warnings for flags and commands

### Commands

```typescript
  static state = 'deprecated'
}
```

```
Warning: The "my command" command has been deprecated.
```

```typescript
  static state = 'deprecated'
  static deprecationOptions = {
    to: 'my other command"
    version: '2.0.0',
  }
}
```

```
Warning: The "my command" command has been deprecated and will be removed in version 2.0.0. Use "my other command" instead.
```

### Flags

```typescript
  static flags = {
    name: Flags.string({
      char: 'n',
      summary: 'Name to print.',
      deprecated: {
        to: 'full-name',
        version: '2.0.0',
      }
    }),
   'full-name': Flags.string({
      summary: 'Name to print.',
    }),
  }
}
```

```
Warning: The "name" flag has been deprecated and will be removed in version 2.0.0. Use "full-name" instead.
```

```typescript
  static flags = {
    name: Flags.string({
      char: 'n',
      summary: 'Name to print.',
      deprecated: true
    }),
  }
}
```

```
Warning: The "name" flag has been deprecated.
```

@W-11836498@

### DIFF
--- a/src/help/util.ts
+++ b/src/help/util.ts
@@ -3,6 +3,7 @@ import {Config as IConfig, HelpOptions} from '../interfaces'
 import {Help, HelpBase} from '.'
 import ModuleLoader from '../module-loader'
 import {collectUsableIds} from '../config/util'
+import {Deprecation} from '../interfaces/parser'
 
 interface HelpBaseDerived {
   new(config: IConfig, opts?: Partial<HelpOptions>): HelpBase;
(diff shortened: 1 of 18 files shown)


src/help/util.ts
```diff
@@ -97,3 +98,32 @@ export function getHelpFlagAdditions(config: IConfig): string[] {
   const additionalHelpFlags = config.pjson.oclif.additionalHelpFlags ?? []
   return [...new Set([...helpFlags, ...additionalHelpFlags]).values()]
 }
+
+export function formatFlagDeprecationWarning(flag: string, opts: true | Deprecation): string {
```
mshanemc, line 102: since there's only 1 word of difference, it could be consolidated?
formatDeprecationWarning(deprecationType: 'command' | 'flag', item: string, opts: true | Deprecation)

mshanemc (dismissed):
1 non-blocking suggestion comment.

can you also put on here what, if anything, needs to be done or should be done for SfdxCommand as a result of this?

Are the deprecations compatible?  Does this get overridden in sfdxCommand?  Should sfdxCommand remove its code for this and rely directly on oclif's ?

mshanemc (approved):

PR #513, opened 2022-10-01
the spread operator doesn't work with getters so we have to manually add it here (same for args)

xlmnxp:
…
fix bug when using tsyringe `@autoInjectable()` decorator the parser cannot read args from class

there 2 fixes for the issue, the first in the this PR and the second to pass args manually
```
const { args, flags } = await this.parse({...Connect, args: Connect.args });
```

maybe the fix is related to  #603

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -253,6 +253,7 @@ export default abstract class Command {
     const opts = {context: this, ...options}
     // the spread operator doesn't work with getters so we have to manually add it here
     opts.flags = options?.flags
+    opts.args = options?.args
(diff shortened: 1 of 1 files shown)


xlmnxp:
> Thanks for the contribution! Before we can merge this, we need @xlmnxp to [sign the Salesforce.com Contributor License Agreement](https://cla.salesforce.com/sign-cla?prUrl=https%3A%2F%2Fgithub.com%2Foclif%2Fcore%2Fpull%2F513).

signed

mdonnalley (approved):

PR #536, opened 2022-10-23
fix: ignore unset environment variables

amcaplan:
When an `env` is set for a boolean flag, if the user doesn't actually specify the value in their ENV, it ends up being treated as though they'd specified `false`. I don't think this is the correct behavior; instead, it should be considered unspecified.

This is especially relevant in 2 cases:

1. Boolean flags with `default: true`. As soon as you add an `env` config, they now effectively default to false.
2. Boolean flags with mutual exclusivity. If you specify one, oclif thinks you've specified both, and the command fails. You can reproduce easily:

```ts

export default class Repro extends Command {
  static description = 'Reproduce exclusive flags bug'

  static flags = {
    exclusive1: Flags.string({
      required: false,
      env: 'SOME_ENV_VAR_1',
      exclusive: ['exclusive2'],
    }),
    exclusive2: Flags.boolean({
      required: false,
      env: 'SOME_ENV_VAR_2',
      exclusive: ['exclusive1'],
    }),
  }

  static args = []

  async run(): Promise<void> {
    const {flags} = await this.parse(Repro)
  }
}
```

```sh
$ bin/dev repro
# everything works fine

$ bin/dev repro --exclusive1
Error: The following error occurred:
  --exclusive1=true cannot also be provided when using --exclusive2
# ... stack trace ...

$ bin/dev repro --exclusive2
Error: The following error occurred:
  --exclusive2=true cannot also be provided when using --exclusive1
# ... stack trace ...
```

### DIFF
--- a/test/parser/parse.test.ts
+++ b/test/parser/parse.test.ts
@@ -910,6 +910,26 @@ See more help with --help`)
           delete process.env.TEST_FOO
         })
       }
+
+      it('ignores unset environment variables', async () => {
+        delete process.env.TEST_FOO
+        const out = await parse([], {
+          flags: {
+            foo: flags.boolean({env: 'TEST_FOO'}),
+          },
+        })
+        expect(out.flags.foo).to.be.undefined
+      })
+
+      it('uses default when environment variable is unset', async () => {
+        delete process.env.TEST_FOO
+        const out = await parse([], {
+          flags: {
+            foo: flags.boolean({env: 'TEST_FOO', default: true}),
+          },
+        })
+        expect(out.flags.foo).to.be.true
+      })
     })
   })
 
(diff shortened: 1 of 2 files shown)


test/parser/parse.test.ts
```diff
@@ -910,6 +910,26 @@ See more help with --help`)
...
+
+      it('uses default when environment variable is unset', async () => {
+        delete process.env.TEST_FOO
+        const out = await parse([], {
+          flags: {
+            foo: flags.boolean({env: 'TEST_FOO', default: true}),
+          },
+        })
+        expect(out.flags.foo).to.be.true
+      })
```
amcaplan, line 932: Just to be 100% clear, both of these tests fail when run against `main`, because in both cases the absent env var is interpreted as falsey:

```
  2 failing

  1) parse
       env
         boolean
           ignores unset environment variables:
     AssertionError: expected false to be undefined
      at Context.<anonymous> (test/parser/parse.test.ts:921:36)

  2) parse
       env
         boolean
           uses default when environment variable is unset:

      AssertionError: expected false to be true
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/parser/parse.test.ts:931:36)
```

mdonnalley (approved):

PR #540, opened 2022-10-28
feat: warn about alias deprecations

mdonnalley:
- Adds `deprecateAliases` to `Command` - allowing you to emit warnings anytime a command alias is used
- Adds `deprecateAliases` to all flags - allowing you to emit warnings anytime a flag alias is used
- Parser is now interprets single character aliases as short char flags


src/command.ts
```diff
@@ -255,10 +260,27 @@ export default abstract class Command {
       if (deprecated) {
         this.warn(formatFlagDeprecationWarning(flag, deprecated))
       }
+
+      const deprecateAliases = this.ctor.flags[flag]?.deprecateAliases
+      const aliases = this.ctor.flags[flag]?.aliases ?? []
+      if (deprecateAliases && aliases) {
```
mshanemc, line 266: `aliases` will always be true since it's `?? []` (empty arrays are truthy).  

Did you maybe want aliases.length?

src/command.ts
```diff
@@ -255,10 +260,27 @@ export default abstract class Command {
...
+        const foundAliases = this.argv.filter(a => aliases.includes(a.replace(/-/g, '')))
```
mshanemc, line 267: what's happening with the `-` ?

src/flags.ts
```diff
@@ -37,7 +37,7 @@ export const help = (opts: Partial<BooleanFlag<boolean>> = {}) => {
     description: 'Show CLI help.',
     ...opts,
     parse: async (_: any, cmd: Command) => {
-      new Help(cmd.config).showHelp(cmd.argv)
+      new Help(cmd.config).showHelp([cmd.id!, ...cmd.argv])
```
mshanemc, line 40: I'm not sure if it matters or not, but I have seen scenarios where a command doesn't have an ID, particularly when we do weird things with them in UT.  

https://github.com/salesforcecli/plugin-source/blob/f5f681b19bcae6a98c25e45e7b461a624f1c2920/test/commands/source/cancel.test.ts#L43

One of those "I'm super cautious about asserting things to TS"

mshanemc (approved):

src/command.ts
mdonnalley: argv has the `--`/`-` on all the flag names e.g., `['--foo', '--bar', '-b']`

But the flag aliases don't have the prefix, e.g. `['foo', 'bar', 'b']`

So in order to find the intersection, we have to remove all the `-`s first

mshanemc: what happens for `target-org` and `no-prompt` and the like?

mshanemc (approved):
Tested with various hyphenated combos of deprecated aliases.

PR #573, opened 2022-12-16
fix: compile linked TS plugins

mshanemc:
oclif/command always registered ts-node.  oclif/core (v1) does not.

In situations where a plugin had no compiled source, oclif was erroring silently when trying to "require" the TS files and using the lib files instead

2 'bugs' emerged from that behavior with linked plugins
* linked TS plugins would produce "not found" errors and required compilation ex https://github.com/forcedotcom/cli/issues/1664 (TS files couldn't be used and lib files didn't exist)
* a plugin developer who compiled their plugin would make aditional changes to the TS source but those files weren't being used unless they recompiled.

this PR attempts to detect linked TS plugins and add a compile step.  **Assumption** it's a linked TS plugins if it's a linked plugin and `typescript` is in its devDependencies.  

If it finds a linked TS plugin, it will recompile the plugin using tsc (--incremental is used to reduce the perf penalty of this).

As a result, for optimal perforance, we recommend that linked plugins are reserved for development time (publish a package!) and to make sure that your package **does** ship its compiled js files and **doesn't** ship its source TS files.

---

Alternatives
1. make sure that the `command` always registered tsnode (and make it a non-dev dep)
2. create an env (or use the one this PR removes) to conditionally tell `command` to register ts-node
3. use an existing one like dev-mode to tell `command` to conditionally register ts-node


mshanemc:
nope this whole approach was a bad idea to fixing the issues

PR #589, opened 2023-01-13
feat: specify flag name in parse error message

MichaelGoberling:
We have use cases where we want to use the `min` and `max` values for an integer flag (See PR [here](https://github.com/adobe/aio-cli-plugin-runtime/pull/270))

It would be nice if the parsing error message included the flag it pertained to

**Before**

```
➜  the-worst-folder git:(main) wow create character --strength=51 --intellect=51 --agility=2
    Error: Expected an integer less than or equal to 50 but received: 51
```

**After**

```
➜  the-best-folder git:(main) wow create character --strength=51 --intellect=51 --agility=2
    Error: Parsing --intellect 
        Expected an integer less than or equal to 50 but received: 51
```

### DIFF
--- a/src/parser/parse.ts
+++ b/src/parser/parse.ts
@@ -177,6 +177,7 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
     return args
   }
 
+  // eslint-disable-next-line complexity
   private async _flags(): Promise<TFlags> {
     const flags = {} as any
     this.metaData.flags = {} as any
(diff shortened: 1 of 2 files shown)


src/parser/parse.ts
```diff
@@ -191,14 +192,24 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
           flags[token.flag] = true
         }
 
-        // eslint-disable-next-line no-await-in-loop
-        flags[token.flag] = await flag.parse(flags[token.flag], this.context, flag)
+        try {
```
mdonnalley, line 195: Could you abstract this out to a function so you don't have to repeat yourself here and on line 206? For instance,

```typescript
function parse(name: string, context: any, flag: BooleanFlag<any> | CompletableOptionFlag<any>) {
  try {
    // return await flag.parse ...
  } catch {
    // throw error
  }
}
```

src/parser/parse.ts
```diff
@@ -191,14 +192,24 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
...
+          // eslint-disable-next-line no-await-in-loop
+          flags[token.flag] = await flag.parse(flags[token.flag], this.context, flag)
+        } catch (error: any) {
+          throw new Error(`Parsing ${token.input} \n\t${error.message}`)
```
mdonnalley, line 199: This should be a `CLIParseError`: https://github.com/oclif/core/blob/main/src/parser/errors.ts#L27

mdonnalley (changes requested):

mdonnalley:
@MichaelGoberling Thanks for the contribution - love the idea! I just made a couple of suggestions to clean up the implementation

src/parser/parse.ts
MichaelGoberling: Absolutely! Just pushed a new commit that uses a new private function 👍

src/parser/parse.ts
MichaelGoberling: Just pushed a change that abstracted this out into a new error type called `FailedFlagParseError`

mdonnalley (approved):

PR #593, opened 2023-01-17
Fixes #592 by directly returning the result of commands to the caller

### LINKED ISSUE #592: Return result in Main.ts run function
**Describe the bug**

Everywhere else throughout the call stack for Commands, the result of a command is propagated upwards - [except here in main.ts](https://github.com/oclif/core/blob/main/src/main.ts#L77). As a result, for those looking to interact with the result of commands (short of commands that offer direct support for the filesystem, which works in the interim as an indirect way to get access to command results), we're left hanging - we can properly await a command, but can't properly resolve the result of a command.

**Expected behavior**

It would be great to update the line from:

```ts
await config.runCommand(id, argvSlice, cmd);
```

to:

```ts
return await config.runCommand(id, argvSlice, cmd);
```

So that we can interact with the result of a command without having to write to the file-system as an intermediary.

 - applicable on any OS
 - applicable as of `@oclif/core` version `1.24.0`

jamessimone:
This change allows callers to interact with the result of commands directly rather than having to rely on the filesystem as an intermediary.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -74,5 +74,5 @@ export async function run(argv = process.argv.slice(2), options?: Interfaces.Loa
   // as an argument, we need to add back the '.' to argv since it was stripped out earlier as part of the
   // command id.
   if (config.pjson.oclif.default === '.' && id === '.' && argv[0] === '.') argvSlice = ['.', ...argvSlice]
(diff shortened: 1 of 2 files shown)


mdonnalley (approved):

jamessimone:
Thank you!

PR #614, opened 2023-02-07
fix: handling quoted strings and options validation for comma-delimited multiple-flag

mshanemc:
1. --multiFlag would validate options agains the combined input BEFORE splitting when it used a delimiter.
2. strings can be double-quoted when they contain spaces.  This removes surrounding quotes from the items before validating against the options.

side note about error messages assertion changes:
test cases would run find from `yarn test` but fail when running the individual test becuase of ASCII color codes in the results.  They've been modified to "include" the expected strings to make them color-blind

### DIFF
--- a/src/parser/parse.ts
+++ b/src/parser/parse.ts
@@ -233,7 +233,7 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
           const values = await Promise.all(
             input.split(flag.delimiter).map(async v => this._parseFlag(v.trim().replace(/^"(.*)"$/, '$1'), flag, token)),
           )
-          // the parse that each element aligns with the `options` property
+          // then parse that each element aligns with the `options` property
           for (const v of values) {
             this._validateOptions(flag, v)
           }
--- a/src/parser/parse.ts
+++ b/src/parser/parse.ts
@@ -231,7 +231,7 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
         if (flag.delimiter && flag.multiple) {
           // split, trim, and remove surrounding doubleQuotes (which would hav been needed if the elements contain spaces)
           const values = await Promise.all(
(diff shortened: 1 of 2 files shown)


test/parser/parse.test.ts
```diff
@@ -351,6 +352,91 @@ See more help with --help`)
         })
         expect(out.flags).to.deep.include({foo: ['a', 'b']})
       })
+      it('allowed options on multiple', async () => {
+        const out = await parse(['--foo', 'a', '--foo=b'], {
+          flags: {
+            foo: Flags.custom({multiple: true, parse: async i => i, options: ['a', 'b']})(),
```
mdonnalley, line 358: Any particular reason for using `Flag.custom` and defining a `parse`? It's not an issue, it just would be more concise to do this:

```
Flags.string({multiple: true, options: ['a', 'b']})
```

mdonnalley (approved):

src/parser/parse.ts
```diff
@@ -227,15 +227,21 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
         flags[token.flag] = await this._parseFlag(flags[token.flag], flag, token)
       } else {
         const input = token.input
-        this._validateOptions(flag, input)
 
         if (flag.delimiter && flag.multiple) {
+          // split, trim, and remove surrounding doubleQuotes (which would hav been needed if the elements contain spaces)
           const values = await Promise.all(
-            input.split(flag.delimiter).map(async v => this._parseFlag(v.trim(), flag, token)),
+            input.split(flag.delimiter).map(async v => this._parseFlag(v.trim().replace(/^"(.*)"$/, '$1'), flag, token)),
```
peternhale, line 234: @mshanemc should this also consider signle quotes, `'a b' 'c d'`?

mshanemc: yeah, I think so.  posix-shells could

mdonnalley (approved):

Issue #616, opened 2023-02-07
NonExistentFlagsError when running this.parse and strict set to false


SferaDev:
**Describe the bug**

In command class inheritance we parse a subset of flags to load specific ones, before V2 we had no problems by passing strict false. With V2 now errors out with NonExistentFlagsError.

**To Reproduce**

Run this.parse with strict set to false, but do not pass required flags.

**Expected behavior**
Error should not be raised.

**Additional context**

PR to fix the issue: https://github.com/oclif/core/pull/606

PR #606, opened 2023-01-31
Avoid nonExistentFlag error with strict set to false

SferaDev:
I'm updating our CLI to v2 and we have noticed that this new check in `Parser.validate` doesn't respect `strict` to false.

Fixes #616

Issue #526, opened 2022-10-17
Adding `strict=false` parses a non-existing flag as a `arg`


iowillhoit:
**Describe the bug**
When you disable argument validation with `static strict = false` it causes non-existing or misspelled flags to be parsed as args. 

**To Reproduce**
1. `static strict = false;` to a command ([docs](https://oclif.io/docs/args#:~:text=For%20variable%20length%20arguments%2C%20disable%20argument%20validation%20with%20static%20strict%20%3D%20false%20on%20the%20command))
2. Parse the arguments on the command and log them to the console
```ts
const {argv} = await this.parse(MyCLI)
console.log('Args:', argv)
```
3. Run the command with a misspelled flag: `bin/dev my:command --hekp`
4. Notice that `--hekp` is parsed as a `arg` instead of getting a warning such as: `Flag '--hekp' does not exist`

**Expected behavior**
I think that it makes the most sense to treat all arguments that start with hyphens (`-` or `--`) as flags even if `strict = false` is used.

**Additional context**
This would very likely be a breaking change. Someone out there may be using `args` that start with hyphens. Let's fix this in `v2`, will add GitHub label.

PR #627, opened 2023-02-13
fix: avoid nonExistentFlag error with strict set to false (#606)

mdonnalley:
Fixes https://github.com/oclif/core/issues/616

### DIFF
--- a/src/parser/validate.ts
+++ b/src/parser/validate.ts
@@ -14,7 +14,7 @@ export async function validate(parse: {
   output: ParserOutput;
 }): Promise<void> {
   function validateArgs() {
-    if (parse.output.nonExistentFlags?.length > 0) {
+    if (parse.input.strict && parse.output.nonExistentFlags?.length > 0) {
       throw new NonExistentFlagsError({parse, flags: parse.output.nonExistentFlags})
     }
 
--- a/test/parser/validate.test.ts
+++ b/test/parser/validate.test.ts
@@ -152,6 +152,62 @@ describe('validate', () => {
     }
   })
 
+  it('throws when extra flags are provided', async () => {
+    const input = {
+      argv: [],
+      flags: {},
+      args: {},
+      strict: true,
+      context: {},
+      '--': true,
+    }
+
+    const output = {
+      args: {},
+      argv: [],
+      flags: {foobar: 'baz'},
+      raw: [],
+      metadata: {
+        flags: {},
+      },
+      nonExistentFlags: ['foobar'],
+    }
+
+    try {
+      // @ts-expect-error
+      await validate({input, output})
+      assert.fail('should have thrown')
+    } catch (error) {
+      const err = error as CLIError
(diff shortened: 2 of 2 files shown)


mdonnalley:
@SferaDev I merged your branch into this one so that we can get the GHA to run for it (we don't allows actions to run on forks for security purposes)

However, in testing this locally I'm starting to question whether we should merge this change. For context, the purpose of that `NonExistentFlagsError` is to pick up on any flags that don't have a corresponding definition in the command, specifically when `strict = false`. The best example of this would be a misspelled flag, e.g. `my-command arg1 --hekp` would throw because `--hekp` is not a defined flag. 

This was all done in order to fix the problem of oclif not knowing about invalid flags when `strict = false` (see this [issue](https://github.com/oclif/core/issues/526)). Before in v1, `--hekp` in the example above would have wrongly been parsed as argument

If we move forward with your change, then we'll essentially be undoing that fix

Could you elaborate on what you're trying to accomplish? Perhaps I can help you figure out a way around this without having to undo any of the fixes we made in v2

SferaDev:
Sure, let me explain where we hit this problem in v2 and was working properly in v1.

In a `BaseCommand` class that we inherit in all other commands we have a utility method to load a global flag.

```ts
async getProfile(): Promise<Profile> {
    const { flags } = await this.parse({ strict: false, flags: { ...BaseCommand.profileFlag } }, this.argv);
    ...
}
```

In the `argv` there will more flags than the ones we know for the command, hence passing the `strict` set to false so that it loads the one that we want but ignores extra ones.

mdonnalley:
@SferaDev so you have a flag that you want all children of `BaseCommand` to inherit?

If that's the case, you define `baseFlags` in your `BaseCommand` that will serve that purpose. Here's the doc for that https://oclif.io/docs/global_flags

You can see a more robust example of how this can be used here: https://oclif.io/docs/base_class

SferaDev:
Let me give a try to that, I don't recall that doc when we implemented this flag so it must be something recent. Will let you know if it's working as it did with the other mechanism.

In any case, with this change you wouldn't be reverting the fix for https://github.com/oclif/core/issues/526, as the problem was that it parsed the incorrect flag as an argument. To me `strict` in parse shouldn't throw, WDYT?

mdonnalley:
@SferaDev `strict = false` is intended to enable an unknown number of arguments to be passed to a command. It's not intended to be a "don't throw any errors" switch.

But, yes, it would revert the fix for #526 since the fix was done specifically for when `strict = false`. oclif has always known how to handle non-existent flags when `strict = true` (or undefined). The problem was that `strict = false` was allowing non-existent flags to be provided, instead of just allowing an unknown number of arguments.

If we merge your fix as it is, `strict = false` would again allow non-existent to be provided (which we don't want to allow)

Let me know if you need any help with implementing `baseFlags` - I'm happy to help

SferaDev:
Base flags worked fine based on the complex example you showed, fine by me closing this as-is

PR #640, opened 2023-02-27
fix: change useable ids to set to improve search perf

peternhale:
Simple change to improve command lookup perf

### DIFF
--- a/src/help/util.ts
+++ b/src/help/util.ts
@@ -41,7 +41,7 @@ function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
   if (argv.length === 1) return argv
 
   const findId = (argv: string[]): string | undefined => {
-    const ids = new Set(collectUsableIds(config.commandIDs))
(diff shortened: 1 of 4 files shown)


src/help/util.ts
```diff
@@ -41,10 +41,10 @@ function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
   if (argv.length === 1) return argv
 
   const findId = (argv: string[]): string | undefined => {
-    const ids = collectUsableIds(config.commandIDs)
+    const ids = new Set(collectUsableIds(config.commandIDs))
```
mdonnalley, line 44: @peternhale would it be better to have `collectUsableIds` return a `Set`? That way other consumers of it get the performance benefits?

mdonnalley (changes requested):

src/help/util.ts
peternhale: I'll take a look, but yes. would also save the sort op in collectUsableIds

mdonnalley (approved):

PR #643, opened 2023-03-03
fix: move `@types/cli-progress` to `devDependencies`

fnesveda:
`@types/cli-progress` was in normal `dependencies`, although all the `@types/xxx` packages are really `devDependencies`.

This was an issue since `@types/cli-progress` depends on `@types/node`, which is huge (7 MB), and it was making our CLI quite fat.

I see this was [added in v2](https://github.com/oclif/core/pull/539), but I feel like it was not on purpose?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
   "author": "Salesforce",
   "bugs": "https://github.com/oclif/core/issues",
   "dependencies": {
-    "@types/cli-progress": "^3.11.0",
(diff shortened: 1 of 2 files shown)


fnesveda:
I'm unable to sign the CLA:
Screenshot 2023-03-03 at 19 41 12

mdonnalley:
Closed by #767

PR #650, opened 2023-03-07
Auto-compile linked plugins

mdonnalley:
Register ts-node for linked plugins so that developers don't have to compile their plugin before running it

More context: https://github.com/forcedotcom/cli/issues/1664
@W-12652231@

### Testing

- `yarn link @oclif/core` in sfdx-cli
- cd to `oclif/plugin-plugins`
  - run `sfdx plugins link`
  - run `rm -rf lib/`
- cd to sfdx-cli
  - [ ] `bin/dev plugins`
  - [ ] `bin/run plugins`
  - [ ] `bin/dev version`
  - [ ] `bin/run version`
- cd to `oclif/plugin-plugins`
  - run `yarn build`
- cd to sfdx-cli
  - [ ] `bin/dev plugins`
  - [ ] `bin/run plugins`
  - [ ] `bin/dev version`
  - [ ] `bin/run version`

### DIFF
--- a/src/config/ts-node.ts
+++ b/src/config/ts-node.ts
@@ -9,6 +9,9 @@ import {Debug} from './util'
 // eslint-disable-next-line new-cap
 const debug = Debug('ts-node')
 
+const TYPE_ROOTS = [`${__dirname}/../node_modules/@types`]
+const ROOT_DIRS: string[] = []
+
 function loadTSConfig(root: string): TSConfig | undefined {
   const tsconfigPath = path.join(root, 'tsconfig.json')
   let typescript: typeof import('typescript') | undefined
@@ -35,17 +38,13 @@ function loadTSConfig(root: string): TSConfig | undefined {
   }
 }
 
-const TS_CONFIGS: Record<string, TSConfig> = {}
-const TYPE_ROOTS = [`${__dirname}/../node_modules/@types`]
-const ROOT_DIRS: string[] = []
 function registerTSNode(root: string) {
-  if (TS_CONFIGS[root]) return
   const tsconfig = loadTSConfig(root)
   if (!tsconfig) return
   debug('registering ts-node at', root)
   const tsNodePath = require.resolve('ts-node', {paths: [root, __dirname]})
   const tsNode: typeof TSNode = require(tsNodePath)
-  TS_CONFIGS[root] = tsconfig
+
   TYPE_ROOTS.push(`${root}/node_modules/@types`)
 
   if (tsconfig.compilerOptions.rootDirs) {
@@ -72,6 +71,7 @@ function registerTSNode(root: string) {
         jsx: 'react',
       },
     })
+    return tsconfig
   } finally {
     process.chdir(cwd)
   }
@@ -86,7 +86,7 @@ export function tsPath(root: string, orig: string, type?: string): string
 export function tsPath(root: string, orig: string | undefined, type?: string): string | undefined
 export function tsPath(root: string, orig: string | undefined, type?: string): string | undefined {
   if (!orig) return orig
-  orig = path.join(root, orig)
+  orig = orig.startsWith(root) ? orig : path.join(root, orig)
 
   const skipTSNode =
     // the CLI specifically turned it off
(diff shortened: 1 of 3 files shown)


src/config/ts-node.ts
```diff
@@ -1,5 +1,6 @@
 import * as fs from 'fs'
 import * as path from 'path'
+import * as TSNode from 'ts-node'
```
cristiand391, line 3: `ts-node` should be moved to the `dependencies` object:
https://github.com/oclif/core/blob/main/package.json#L76

src/config/ts-node.ts
```diff
@@ -8,6 +9,9 @@ import {Debug} from './util'
 // eslint-disable-next-line new-cap
 const debug = Debug('ts-node')
 
+const TYPE_ROOTS = [`${__dirname}/../node_modules/@types`]
```
cristiand391, line 12: I see these paths are used by `ts-node`, will this work on Windows too or should it use `path.join|sep`?

mdonnalley: I think so. This was taken from [oclif/config](https://github.com/oclif/config/blob/main/src/ts-node.ts#L50) which as far I know never had windows path issues

src/config/ts-node.ts
mdonnalley: I can move it but I'm not sure that it's necessary since we're only importing it for a type. 

If you compile and look at this file there's no `ts-node` import,
```javascript
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.tsPath = void 0;
const fs = require("fs");
const path = require("path");
const settings_1 = require("../settings");
const util_1 = require("../util");
const util_2 = require("./util");
// eslint-disable-next-line new-cap
const debug = (0, util_2.Debug)('ts-node');
```

cristiand391: this `require.resolve` will try to look up the location of `ts-node` at runtime.

compiled output:
```js
function registerTSNode(root) {
    const tsconfig = loadTSConfig(root);
    if (!tsconfig)
        return;
    debug('registering ts-node at', root);
    const tsNodePath = require.resolve('ts-node', { paths: [root, __dirname] });
...
```

https://github.com/oclif/core/pull/650/files/83c60b3b64534ccae2a3c3251d3b492f5fc383e4#diff-5622386b4043e10ee71bec48dcb5cefd6c9cae20cecc1124bd40d82ac82ab8b6R45

cristiand391 (approved):

cristiand391:
**QA notes**:

✅ plugin compiled by `ts-node` (`rmrf lib`, updated command, see update after running it)
✅ JS plugin (no tsconfig file), running command just throws not found.

ts-node:
I added some code to print paths in `oclif/core` to see what ts-node was loading it, see results:

1) `ts-node` as plugin devDep
`plugin-commands` has `ts-node` as a [devDep](https://github.com/oclif/plugin-commands/blob/main/package.json#L29.
when loading ts-node, oclif uses the `ts-node` in the plugin's node_modules folder:
```
➜  sfdx-cli git:(main) ✗ ./bin/run commands
ROOT: /Users/cdominguez/code/gh/oclif/plugin-commands
__dirname: /Users/cdominguez/code/gh/oclif/core/lib/config
tsNodePath: /Users/cdominguez/code/gh/oclif/plugin-commands/node_modules/ts-node/dist/index.js
 Command                                    Summary
...
```

2) plugin doesn't have `ts-node` dep
For this I used `plugin-plugins`, removed `ts-node` from devDeps and then linked it.
oclif loads `ts-node` from `@oclif/core`:
```
➜  sfdx-cli git:(main) ✗ ./bin/run plugins
ROOT: /Users/cdominguez/code/gh/oclif/plugin-plugins
__dirname: /Users/cdominguez/code/gh/oclif/core/lib/config
tsNodePath: /Users/cdominguez/code/gh/oclif/core/node_modules/ts-node/dist/index.js
@oclif/plugin-plugins 2.3.2 (link) /Users/cdominguez/code/gh/oclif/plugin-plugins
```

PR #659, opened 2023-03-15
fix: pass plugin type when resolving a module path

cristiand391:
This fixes a bug where linked plugins with hooks were causing a CLI to fail with the `ModuleLoadError: [MODULE_NOT_FOUND] require failed to load`.

https://github.com/oclif/core/pull/650 introduced compilation at runtime for linked TS plugins using `ts-node`.
When trying to compile hooks from linked plugins oclif isn't passing the `plugin.type` prop to the ts path resolver so `tsPath` was just skipping the compilation step.

## Before:
Screenshot 2023-03-16 at 10 07 02

## After
Screenshot 2023-03-16 at 10 07 32

### Testing

**Repro**
- cd to `plugin-info` (doctor hook)
  - run `rm -rf lib`
  - run `../sfdx-cli/bin/run plugins:link .`
  - run `../sfdx-cli/bin/run`, should see the `MODULE_NOT_FOUND`

**QA**
  - yarn link @oclif/core in sfdx-cli
  - cd to `plugin-info` (doctor hook)
  - run `rm -rf lib`
  - run `../sfdx-cli/bin/run plugins:link .`
  - run `../sfdx-cli/bin/run`, CLI should work.

### DIFF
--- a/src/module-loader.ts
+++ b/src/module-loader.ts
@@ -136,11 +136,15 @@ export default class ModuleLoader {
     let isESM: boolean
     let filePath: string
 
+    const isPlugin = (config: IConfig|IPlugin): config is IPlugin => {
+      return (<IPlugin>config).type !== undefined
+    }
+
     try {
       filePath = require.resolve(modulePath)
       isESM = ModuleLoader.isPathModule(filePath)
     } catch {
-      filePath = Config.tsPath(config.root, modulePath)
+      filePath = isPlugin(config) ? Config.tsPath(config.root, modulePath, config.type) : Config.tsPath(config.root, modulePath)
 
       let fileExists = false
       let isDirectory = false


src/module-loader.ts
```diff
@@ -136,11 +136,15 @@ export default class ModuleLoader {
...
+    const isPlugin = (config: IConfig|IPlugin): config is IPlugin => {
+      return (<IPlugin>config).type !== undefined
+    }
+
     try {
       filePath = require.resolve(modulePath)
       isESM = ModuleLoader.isPathModule(filePath)
     } catch {
-      filePath = Config.tsPath(config.root, modulePath)
+      filePath = isPlugin(config) ? Config.tsPath(config.root, modulePath, config.type) : Config.tsPath(config.root, modulePath)
```
cristiand391, line 147: pass `plugin.type` to `Config.tsPath` if `config` is a plugin.

hooks get loaded here:
https://github.com/oclif/core/blob/main/src/config/config.ts#L310

usually these are paths to compiled JS files like this one:
https://github.com/salesforcecli/plugin-info/blob/main/package.json#L93

Linked plugins are supposed to get compiled at runtime by ts-node but hooks were being skipped because the `plugin.type` wasn't being passed to `tsPath`:
https://github.com/oclif/core/blob/main/src/config/ts-node.ts#L98

commands from linked plugins are loaded here (see `plugin.type` is being passed):
https://github.com/oclif/core/blob/main/src/config/plugin.ts#L182

src/module-loader.ts
```diff
@@ -136,11 +136,15 @@ export default class ModuleLoader {
     let isESM: boolean
     let filePath: string
 
+    const isPlugin = (config: IConfig|IPlugin): config is IPlugin => {
```
cristiand391, line 139: TS guard, used to pass `plugin.type` when config comes from a plugin.

mshanemc:
QA: followed steps above setup steps above.  

Modify linked plugin (doctor) and reran, change runs from TS with no compilation required.

Screenshot 2023-03-16 at 8 39 26 AM

modified the hook, change also reflected immediately
Screenshot 2023-03-16 at 8 42 56 AM

mshanemc (approved):

PR #665, opened 2023-03-20
feat: add param noSensitiveData to Plugin.load

peternhale:
@W-12513575@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -813,11 +813,11 @@ export class Config implements IConfig {
 }
 
 // when no manifest exists, the default is calculated.  This may throw, so we need to catch it
-const defaultFlagToCached = async (flag: CompletableOptionFlag<any>, noSensitiveData = false) => {
+const defaultFlagToCached = async (flag: CompletableOptionFlag<any>, suppressSensitiveData = false) => {
   // Prefer the helpDefaultValue function (returns a friendly string for complex types)
   if (typeof flag.defaultHelp === 'function') {
     try {
-      return await flag.defaultHelp({options: flag, flags: {}}, noSensitiveData)
+      return await flag.defaultHelp({options: flag, flags: {}}, suppressSensitiveData)
     } catch {
       return
     }
@@ -826,18 +826,18 @@ const defaultFlagToCached = async (flag: CompletableOptionFlag<any>, noSensitive
   // if not specified, try the default function
   if (typeof flag.default === 'function') {
     try {
-      return await flag.default({options: {}, flags: {}}, noSensitiveData)
+      return await flag.default({options: {}, flags: {}}, suppressSensitiveData)
     } catch {}
   } else {
     return flag.default
   }
 }
 
-const defaultArgToCached = async (arg: Arg<any>, noSensitiveData = false): Promise<any> => {
+const defaultArgToCached = async (arg: Arg<any>, suppressSensitiveData = false): Promise<any> => {
   // Prefer the helpDefaultValue function (returns a friendly string for complex types)
   if (typeof arg.defaultHelp === 'function') {
     try {
-      return await arg.defaultHelp({options: arg, flags: {}}, noSensitiveData)
+      return await arg.defaultHelp({options: arg, flags: {}}, suppressSensitiveData)
     } catch {
       return
     }
@@ -846,14 +846,14 @@ const defaultArgToCached = async (arg: Arg<any>, noSensitiveData = false): Promi
   // if not specified, try the default function
   if (typeof arg.default === 'function') {
     try {
-      return await arg.default({options: arg, flags: {}}, noSensitiveData)
+      return await arg.default({options: arg, flags: {}}, suppressSensitiveData)
     } catch {}
   } else {
     return arg.default
   }
 }
 
-export async function toCached(c: Command.Class, plugin?: IPlugin | undefined, noSensitiveData?: boolean): Promise<Command.Cached> {
+export async function toCached(c: Command.Class, plugin?: IPlugin | undefined, suppressSensitiveData?: boolean): Promise<Command.Cached> {
   const flags = {} as {[k: string]: Command.Flag.Cached}
 
   for (const [name, flag] of Object.entries(c.flags || {})) {
@@ -894,7 +894,7 @@ export async function toCached(c: Command.Class, plugin?: IPlugin | undefined, n
         dependsOn: flag.dependsOn,
         relationships: flag.relationships,
         exclusive: flag.exclusive,
-        default: await defaultFlagToCached(flag, noSensitiveData),
+        default: await defaultFlagToCached(flag, suppressSensitiveData),
         deprecated: flag.deprecated,
         deprecateAliases: c.deprecateAliases,
         aliases: flag.aliases,
@@ -914,7 +914,7 @@ export async function toCached(c: Command.Class, plugin?: IPlugin | undefined, n
       description: arg.description,
       required: arg.required,
       options: arg.options,
-      default: await defaultArgToCached(arg, noSensitiveData),
+      default: await defaultArgToCached(arg, suppressSensitiveData),
       hidden: arg.hidden,
     }
   }
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -833,11 +833,11 @@ const defaultFlagToCached = async (flag: CompletableOptionFlag<any>, noSensitive
   }
 }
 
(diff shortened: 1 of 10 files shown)


src/config/config.ts
```diff
@@ -813,11 +813,11 @@ export class Config implements IConfig {
 }
 
 // when no manifest exists, the default is calculated.  This may throw, so we need to catch it
-const defaultFlagToCached = async (flag: CompletableOptionFlag<any>) => {
+const defaultFlagToCached = async (flag: CompletableOptionFlag<any>, noSensitiveData = false) => {
```
peternhale, line 816: add noSensitiveData param to defaultFlagToCached. Value is pass to defaultHelp and default

src/config/config.ts
```diff
@@ -826,18 +826,18 @@ const defaultFlagToCached = async (flag: CompletableOptionFlag<any>) => {
...
-      return await flag.default({options: {}, flags: {}})
+      return await flag.default({options: {}, flags: {}}, noSensitiveData)
     } catch {}
   } else {
     return flag.default
   }
 }
 
-const defaultArgToCached = async (arg: Arg<any>): Promise<any> => {
+const defaultArgToCached = async (arg: Arg<any>, noSensitiveData = false): Promise<any> => {
```
peternhale, line 836: add noSensitiveData param to defaultArgToCached. Value is pass to defaultHelp and default

src/config/config.ts
```diff
@@ -846,14 +846,14 @@ const defaultArgToCached = async (arg: Arg<any>): Promise<any> => {
...
-      return await arg.default({options: {}, flags: {}})
+      return await arg.default({options: arg, flags: {}}, noSensitiveData)
     } catch {}
   } else {
     return arg.default
   }
 }
 
-export async function toCached(c: Command.Class, plugin?: IPlugin): Promise<Command.Cached> {
+export async function toCached(c: Command.Class, plugin?: IPlugin | undefined, noSensitiveData?: boolean): Promise<Command.Cached> {
```
peternhale, line 856: add noSensitiveData param to toCached. Value will be passes to defaultFlagToCached and defaultArgToCached functions

src/config/plugin.ts
```diff
@@ -136,7 +136,7 @@ export class Plugin implements IPlugin {
 
   constructor(public options: PluginOptions) {}
 
-  public async load(): Promise<void> {
+  public async load(noSensitiveData?: boolean): Promise<void> {
```
peternhale, line 139: add noSensitiveData param to load. Value is passed to _manifest

src/config/plugin.ts
```diff
@@ -279,7 +279,7 @@ export class Plugin implements IPlugin {
       version: this.version,
       commands: (await Promise.all(this.commandIDs.map(async id => {
         try {
-          return [id, await toCached(await this.findCommand(id, {must: true}), this)]
+          return [id, await toCached(await this.findCommand(id, {must: true}), this, noSensitiveData)]
```
peternhale, line 282: Pass noSensitiveData to toCached

src/interfaces/parser.ts
```diff
@@ -55,11 +55,11 @@ export type DefaultContext<T> = {
   flags: Record<string, string>;
 }
 
-export type FlagDefault<T, P = CustomOptions> = T | ((context: DefaultContext<P & OptionFlag<T, P>>) => Promise<T>)
-export type FlagDefaultHelp<T, P = CustomOptions> = T | ((context: DefaultContext<P & OptionFlag<T, P>>) => Promise<string | undefined>)
+export type FlagDefault<T, P = CustomOptions> = T | ((context: DefaultContext<P & OptionFlag<T, P>>, noSensitiveData?: boolean) => Promise<T>)
```
peternhale, line 58: Add noSensitiveData to flag and arg default function types

test/help/help-test-utils.ts
```diff
@@ -47,7 +47,7 @@ export class TestHelp extends Help {
 
 export const commandHelp = (command?: any) => ({
   async run(ctx: {help: TestHelp; commandHelp: string; expectation: string}) {
-    const cached = await toCached(command!, {} as any)
+    const cached = await toCached(command!, {} as any, false)
```
peternhale, line 50: Command help is being called, do not suppress sensitive data

peternhale:
@mshanemc With this change to core, oclif manifest will be changed to pass true to plugin.load(). Any custom default or defaultHelp functions will then be able to omit sensitive info from return values.

mshanemc (changes requested):
So we pass it around like a hot potato to make sure the various "consumers" get the correct value, but it's up to the user-defined `default` function to decide what to do.

I don't love the name...would I set it to true if I'm including sensitive information that I want omitted (and therefore `containsSensitiveInformation`:true would be a better name.

Or...does `noSensitiveData : true` mean "this contains no sensitive data" and so things should proceed as normal?

peternhale:
> So we pass it around like a hot potato to make sure the various "consumers" get the correct value, but it's up to the user-defined `default` function to decide what to do.
> 
> I don't love the name...would I set it to true if I'm including sensitive information that I want omitted (and therefore `containsSensitiveInformation`:true would be a better name.
> 
> Or...does `noSensitiveData : true` mean "this contains no sensitive data" and so things should proceed as normal?

Backwards compatibility. Plugin.load is currently a no parameter function. I could change the name to `suppressSensitiveData` with default of false.

src/interfaces/parser.ts
```diff
@@ -43,6 +43,7 @@ export type Metadata = {
 
 type MetadataFlag = {
   setFromDefault?: boolean;
+  defaultHelp?: unknown;
```
peternhale, line 46: Need to retain value from defaultHelp to be used in later steps of parse/validation.

src/interfaces/parser.ts
```diff
@@ -208,7 +292,7 @@ export type BooleanFlag<T> = FlagProps & BooleanFlagProps & {
 
 export type OptionFlagDefaults<T, P = CustomOptions, M = false> = FlagProps & OptionFlagProps & {
   parse: FlagParser<T, string, P>
-  defaultHelp?: FlagDefaultHelp<T>;
+  defaultHelp?: T | FlagDefaultHelp<T>;
```
peternhale, line 295: I need to check this change. Should not be needed

src/parser/parse.ts
```diff
@@ -271,7 +279,7 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
       }
 
       if (!(k in flags) && flag.default !== undefined) {
-        this.metaData.flags[k] = {setFromDefault: true}
+        this.metaData.flags[k] = {...this.metaData.flags[k], setFromDefault: true}
```
peternhale, line 282: make sure all metadata captured are retained

src/parser/validate.ts
```diff
@@ -118,7 +118,8 @@ export async function validate(parse: {
       if (parse.output.metadata.flags && parse.output.metadata.flags[name]?.setFromDefault)
         continue
       if (parse.output.flags[flag] !== undefined) {
-        return {...base, status: 'failed', reason: `--${flag}=${parse.output.flags[flag]} cannot also be provided when using --${name}`}
+        const flagValue = parse.output.metadata.flags?.[flag]?.defaultHelp ?? parse.output.flags[flag]
```
peternhale, line 121: Exclusive use error to be created here. To avoid `--target-org=[object Object] cannot also be provided when using --all`, see if defaultHelp value is present and if so use that in message.

cristiand391 (approved):

src/config/plugin.ts
```diff
@@ -136,7 +136,13 @@ export class Plugin implements IPlugin {
...
   constructor(public options: PluginOptions) {}
 
-  public async load(): Promise<void> {
+  /**
+   * Loads a plugin
+   * @param isWritingManifest - if true, exclude selected data from manifest
+   * default is false to maintain backwards compatibility
+   * @returns Promise<void>
+   */
+  public async load(isWritingManifest?: boolean): Promise<void> {
```
cristiand391, line 145: Have you considered adding `isWritingManifest` as a new field in the `options` object here?
https://github.com/oclif/core/blob/main/src/config/plugin.ts#L137

That way we could load plugins in `oclif manifest` like this:
```ts
const plugin = new Plugin({root: fullPath, isWritingManifest: true})
await plugin.load()
```

instead of adding a new arg to `Plugin.load`

PR #678, opened 2023-03-29
fix: default help behaves properly

peternhale:
@W-12513575@

### DIFF
--- a/src/parser/parse.ts
+++ b/src/parser/parse.ts
@@ -279,12 +279,14 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
         }) : flag.default)
         flags[k] = defaultValue
       }
+    }
 
-      if ((k in flags) && Reflect.has(flag, 'defaultHelp')) {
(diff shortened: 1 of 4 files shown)


src/parser/parse.ts
```diff
@@ -284,11 +273,29 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
...
+        const defaultValue = (typeof flag.default === 'function' ? await flag.default({
+          options: flag,
+          flags, ...this.context,
+        }) : flag.default)
         flags[k] = defaultValue
       }
     }
 
+    for (const k of Object.keys(this.input.flags)) {
+      if ((k in flags) && Reflect.has(this.input.flags[k], 'defaultHelp')) {
```
cristiand391, line 285: could this be done in the previous for loop? both iterate over the keys of `this.input.flags`
https://github.com/oclif/core/blob/main/src/parser/parse.ts#L268-L290

cristiand391 (approved):

PR #683, opened 2023-04-04
fix: update config to not use node prefix imports

jondmcelroy:
- to support node >=14
 - https://nodejs.org/docs/latest-v14.x/api/esm.html#esm_node_imports
 - only supported  v14.18.0 (ESM import and CommonJS require()), v14.13.1 (only ESM import)


jondmcelroy:
I signed the CLA but not sure how to get it to re run.

mdonnalley:
@jondmcelroy v3 dropped support for node 14 so there's no longer a need to remove the `node:` prefix. I'm going to close this PR but please feel free to reopen if you think there's more here that needs to be addressed. Thanks!

PR #687, opened 2023-04-11
fix:  jsonEnabled not handling after pass-through

peternhale:
@W-13012400@

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -125,6 +125,8 @@ export abstract class Command {
 
   public static hasDynamicHelp = false
 
+  protected static '_--' = false
+
   protected static _enableJsonFlag = false
 
   public static get enableJsonFlag(): boolean {
@@ -140,6 +142,18 @@ export abstract class Command {
       this.flags = {} // force the flags setter to run
       delete this.flags?.json
     }
+  }
+
+  public static get '--'(): boolean {
+    return Command['_--']
+  }
+
+  public static set '--'(value: boolean) {
+    Command['_--'] = value
+  }
+
+  public get passThroughEnabled(): boolean {
+    return Command['_--']
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/command.ts
```diff
@@ -259,8 +273,22 @@ export abstract class Command {
...
+   *
+   * @returns {boolean} true if the command supports json and the --json flag is present
+   */
   public jsonEnabled(): boolean {
-    return this.ctor.enableJsonFlag && this.argv.includes('--json')
+    // if the command doesn't support json, return false
+    if (!this.ctor.enableJsonFlag) return false
+    // if the command parameter pass through is enabled, return true if the --json flag is before the '--' separator
+    if (this.passThroughEnabled) {
+      const ptIndex = this.argv.indexOf('--')
```
mshanemc, line 286: wouldn't this pick up the first `--anything`, like a regular flag ?

mshanemc:
```suggestion
      const ptIndex = this.argv.indexOf(' -- ')
```

test/command/command.test.ts
```diff
@@ -449,4 +449,90 @@ describe('command', () => {
...
+    .it('json enabled/pass through enabled/--json flag before --/jsonEnabled() should be true')
+
+    fancy
+    .stdout()
+    .do(async () => {
+      class CMD extends Command {
+          static enableJsonFlag = true
+          static '--' = true
+          async run() {
+            const {flags} = await cmd.parse(CMD, ['--', '--json'])
```
mshanemc, line 502: maybe have a test case that has some other long/short flags before the  ` -- ` to verify the behavior

test/interfaces/flags.test.ts
```diff
@@ -126,6 +126,8 @@ class MyCommand extends BaseCommand {
     defaultArrayFlag: arrayFlag({default: ['foo', 'bar']}),
   }
 
+  public static '--' = true
```
mshanemc, line 129: What's that doing?

mshanemc (changes requested):
couple questions.  I think the indexOf('--') is a bug

src/command.ts
peternhale: @mshanemc no it will not. `this.argv` at this moment is an array of arguments, not a string, so this is finding the index of `--`.

test/interfaces/flags.test.ts
peternhale: It is enabling passthrough. I can add a comment

test/command/command.test.ts
peternhale: Will have a look.

test/interfaces/flags.test.ts
mshanemc: If it weren't there, what would be different and why?

test/command/command.test.ts
mshanemc: I did a few locally to check. Looks! good

src/command.ts
mshanemc: yeah, you're right.  It works in all the ways I tried it

test/interfaces/flags.test.ts
mshanemc: asking because removing it didn't make any noticeable different, and no tests were referring to it.

mshanemc (approved):

PR #691, opened 2023-04-20
fix: pass flag to default function

peternhale:
@W-11868418@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -826,7 +826,7 @@ const defaultFlagToCached = async (flag: CompletableOptionFlag<any>, isWritingMa
   // if not specified, try the default function
   if (typeof flag.default === 'function') {
     try {
-      return await flag.default({options: {}, flags: {}}, isWritingManifest)
+      return await flag.default({options: flag, flags: {}}, isWritingManifest)
     } catch {}
   } else {
     return flag.default
@@ -846,7 +846,7 @@ const defaultArgToCached = async (arg: Arg<any>, isWritingManifest = false): Pro
   // if not specified, try the default function
   if (typeof arg.default === 'function') {
     try {
-      return await arg.default({options: arg, flags: {}}, isWritingManifest)
+      return await arg.default(isWritingManifest)
     } catch {}
   } else {
     return arg.default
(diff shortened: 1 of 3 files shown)


src/config/config.ts
```diff
@@ -826,7 +826,7 @@ const defaultFlagToCached = async (flag: CompletableOptionFlag<any>, isWritingMa
   // if not specified, try the default function
   if (typeof flag.default === 'function') {
     try {
-      return await flag.default({options: {}, flags: {}}, isWritingManifest)
+      return await flag.default({options: flag, flags: {}}, isWritingManifest)
```
peternhale, line 829: Need to pass the flag definition to default function to allow the function to access options from the flag

src/parser/parse.ts
```diff
@@ -273,10 +273,7 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
 
       if (!(k in flags) && flag.default !== undefined) {
         this.metaData.flags[k] = {...this.metaData.flags[k], setFromDefault: true}
-        const defaultValue = (typeof flag.default === 'function' ? await flag.default({
-          options: flag,
-          flags, ...this.context,
-        }) : flag.default)
+        const defaultValue = (typeof flag.default === 'function' ? await flag.default({options: flag, flags}) : flag.default)
```
peternhale, line 276: Need to pass flag definition to default function, along with current set of realized flags.

cristiand391 (commented):
The WI mentions `sf env create scratch --help` doesn't show the default values `--wait and --duration-days`, which don't use a default function for these values. Is there any follow up work for it to work? 

https://github.com/salesforcecli/plugin-org/blob/main/src/commands/org/create/scratch.ts#L86-L103

peternhale:
> The WI mentions `sf env create scratch --help` doesn't show the default values `--wait and --duration-days`, which don't use a default function for these values. Is there any follow up work for it to work?
> 
> https://github.com/salesforcecli/plugin-org/blob/main/src/commands/org/create/scratch.ts#L86-L103

I tested plugin-org locally with `@salesforce/sf-plugins-core": "2.2.10-beta.1`

cristiand391:
@peternhale QA looks good (tested with org commands, wait default and and devhub flags (were working before, tested to confirm it didn't break).

e2e test fail on latest node (linux and windows), I re-ran the jobs yesterday and also tested locally with node v20 and can repro the failure.

peternhale:
> @peternhale QA looks good (tested with org commands, wait default and and devhub flags (were working before, tested to confirm it didn't break).
> 
> e2e test fail on latest node (linux and windows), I re-ran the jobs yesterday and also tested locally with node v20 and can repro the failure.

I'll install 20 and see what is up.
Thank you

PR #708, opened 2023-06-06
fix: don't truncate at OCLIF level

WillieRuemmele:
What does this PR do?
no-truncate table output at the `oclif` level

What issues does this PR fix or reference?
@W-13003428@
https://github.com/forcedotcom/cli/issues/2048

for when commands call directly into OCLIF's `ux.table` method don't truncate output

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "lint": "eslint . --ext .ts --config .eslintrc",
     "posttest": "yarn lint",
     "prepack": "yarn run build",
-    "test": "mocha --forbid-only \"test/**/*.test.ts\"",
+    "test": "mocha --forbid-only \"test/**/*.test.ts\" --tiimeout 1200000",
     "test:e2e": "mocha --forbid-only \"test/**/*.e2e.ts\" --timeout 1200000",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -109,9 +109,9 @@
     "lint": "eslint . --ext .ts --config .eslintrc",
     "posttest": "yarn lint",
     "prepack": "yarn run build",
-    "test": "mocha --forbid-only \"test/**/*.test.ts\"",
+    "test": "mocha --forbid-only \"test/**/*.test.ts\" --tiimeout 1200000",
```
cristiand391, line 112: typo: `--tiimeout`

src/cli-ux/styled/table.ts
```diff
@@ -42,7 +42,7 @@ class Table<T extends Record<string, unknown>> {
       extended,
       filter,
       'no-header': options['no-header'] ?? false,
-      'no-truncate': options['no-truncate'] ?? false,
```
cristiand391, line 45: any reason not to do this in `sf-plugins-core/sfCommand`?

this will affect other oclif CLIs (commands showing sensitive info that was truncated before, CLI scripts parsing tables, etc) and should be a major release.

PR #713, opened 2023-06-15
fix: correctly load legacy plugins

cristiand391:
This PR fixes an issue where CLIs that includes `@oclif/plugin-legacy` and at least one [old, invalid plugin ](https://github.com/oclif/core/blob/main/src/config/plugin.ts#L161-L165) would get legacy plugins as the only ones installed.

 left: ci-v5 plugin installed, shows as the only plugin (`--core` should show core plugins too), `plugins:inspect webhooks` fails even though `@heroku-cli/plugin-webhooks` is a core plugin. 
right: with oclif fix, `plugins:inspect` is able to find it.
Screenshot 2023-06-15 at 13 31 52

plugin-legacy init hook where invalid, legacy plugins are converted:
https://github.com/oclif/plugin-legacy/blob/main/src/hooks/init.ts

### Repro

1. install Heroku CLI: `npm i -g heroku`
2. check core plugins: `heroku plugins --core`, this only returns `addons-v5 8.1.7 (core)` because it's the first invalid plugin found.
3. modify the splice line in the JS code 
4. run `heroku plugins --core` again, should get all core plugins.

[skip-validate-pr]

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -805,10 +805,24 @@ export class Config implements IConfig {
     return commandPlugins[0]
   }
 
+  /**
+    * Insert legacy plugins
+    *
+    * Replace invalid CLI plugins (cli-engine plugins, mostly Heroku) loaded via `this.loadPlugins`
+    * with oclif-compatible ones returned by @oclif/plugin-legacy init hook.
+    *
+    * @param plugins array of oclif-compatible plugins
+    * @returns void
+    */
   private insertLegacyPlugins(plugins: IPlugin[]) {
     for (const plugin of plugins) {
       const idx = this.plugins.findIndex(p => p.name === plugin.name)
-      if (idx !== -1) this.plugins = this.plugins.splice(idx, 1, plugin)
+      if (idx !== -1) {
+        // invalid plugin instance found in `this.plugins`
+        // replace with the oclif-compatible one
+        this.plugins.splice(idx, 1, plugin)
+      }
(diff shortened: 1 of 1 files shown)


src/config/config.ts
```diff
@@ -805,10 +805,24 @@ export class Config implements IConfig {
...
+    * Replace invalid CLI plugins (cli-engine plugins, mostly Heroku) loaded via `this.loadPlugins`
+    * with oclif-compatible ones returned by @oclif/plugin-legacy init hook.
+    *
+    * @param plugins array of oclif-compatible plugins
+    * @returns void
+    */
   private insertLegacyPlugins(plugins: IPlugin[]) {
     for (const plugin of plugins) {
       const idx = this.plugins.findIndex(p => p.name === plugin.name)
-      if (idx !== -1) this.plugins = this.plugins.splice(idx, 1, plugin)
```
cristiand391, line 811: `Array.splice` mutates the array and returns an array of deleted items in this case, see:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice#return_value

so, `this.plugins` was set to the non-compatible version of the plugin loaded here:
https://github.com/oclif/core/blob/a2efc040855969678a6aaf14d3ceb1d25d7a8411/src/config/config.ts#L193

k80bowman (approved):
Great find, thank you!

PR #717, opened 2023-06-22
[DO NOT MERGE] Adds a new subtype for non-boolean types

iowillhoit:
Right now, a oclif flag `type` can only be `boolean` or `option`. This limits the information that you could learn about the command programmatically using the `--json` flag. Rather than adding other flag types for `type`, we add a new `subType` value that specifies the actual flag type

https://github.com/forcedotcom/cli/issues/279

Testing
- Pull branch
- `yarn build`
- `yarn link`
- Go to `plugin-command`
- yarn link "@oclif/core"
- Run `bin/dev commands --json`
- Note that the new `subType` values are correct for this command's flags

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -416,6 +416,7 @@ export namespace Command {
     examples?: Example[];
     strict?: boolean;
     type?: string;
+    subType?: string;
     pluginName?: string;
     pluginType?: string;
     pluginAlias?: string;
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -897,6 +897,7 @@ export async function toCached(c: Command.Class, plugin?: IPlugin | undefined, i
       flags[name] = {
         name,
         type: flag.type,
+        subType: flag.subType,
         char: flag.char,
         summary: flag.summary,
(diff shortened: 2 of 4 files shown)


mdonnalley:
@iowillhoit I'm not convinced yet that this is something we can actually support. While we can put a `subType` on oclif's native flag types, it doesn't solve the problem of custom flags created by other people. How do ensure that everyone adds a `subType` to their custom flags? We could make it a required property but then you're looking a major version bump.

Instead you have to look at ways for `custom` to pull that information out for every single flag, which I'm stumped on. The type of flag is really the type it returns from its `parse` - for instance, `Flags.string` is a `string` and `Flags.duration` is a `Duration`. Those types are the [first type parameter](https://github.com/oclif/core/blob/main/src/flags.ts#L31) given to `custom` but we don't have that type information available to us at runtime. So I'm not seeing at the moment how we can actually provide this information universally for any flag

iowillhoit:
That makes sense to me, going to close the PR and Issue
cc @mshanemc

PR #721, opened 2023-06-27
Enforced arg prefixes (-) when matching flags.

mfreydl:
Spent some time diagnosing the issue that turned out to be from Willie's code 
[ca9fe38](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)
Here's a fix to it that will enforce the prefixes "--" on the flag values and prevent his false positives, though I can see you reverted his commit already.  
Anyway, thought this might still be useful.  My scenario is reflected in the `parse.test.ts` unit test `doesn\'t throw when 2nd char in value matches a flag char`

### DIFF
--- a/src/parser/parse.ts
+++ b/src/parser/parse.ts
@@ -110,9 +110,16 @@ export class Parser<T extends ParserInput, TFlags extends OutputFlags<T['flags']
       return Object.keys(this.input.flags).find(k => this.input.flags[k].char === char)
     }
 
+
+    const findFlag = (arg: string): [string|undefined,boolean] => {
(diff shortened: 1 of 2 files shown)


WillieRuemmele:
Hey @mfreydl - you're too fast and too good 🏆 

we've added integration tests to make sure we won't break something like this again. I've added this to the new flag validation PR that was merged in today

Thank you :D

mfreydl:
Nice! oclif's been a great product for us - happy to help!

PR #722, opened 2023-06-29
Properly truncate table cells that contain fullwidth characters or ANSI escape sequences

andreacorbellini:
The [current implementation of `outputTable`](https://github.com/oclif/core/blob/fbdf25b9eff4c7b8e3dd36ec29a6e035ce2ce5a6/src/cli-ux/styled/table.ts#L283-L288) is correctly computing cell widths using `sw()`, which takes into account fullwidth characters and ANSI escape sequences:

```
const visualWidth = sw(d)
```

However, when truncating the cell, the raw string is getting sliced without considering any special characters:

```
cell = cell.slice(0, width - 2) + '… '
```

Because of that, a red-colored string like `"\u001b[31m hello \u001b[39m"` may get incorrectly truncated to `"\u001b[31m hell… "`, which will make the red color spill over adjacent cells. Similarly, a string containing fullwidth characters like `"你好我是意大利人"` may get truncated to twice the expected width, resulting in misaligned output.

This commit fixes the problem by using [slice-ansi](https://www.npmjs.com/package/slice-ansi) instead of `slice()`.


WillieRuemmele:
Hi @andreacorbellini we'll be happy to get this fixed. Can this be done with adding extra dependencies?

package.json
```diff
@@ -68,6 +69,7 @@
     "fancy-test": "^2.0.16",
     "globby": "^11.1.0",
     "husky": "6",
+    "lodash": "^4.17.21",
```
shetzel, line 72: We'd like to avoid adding libraries when not necessary, even dev deps.  Can you please change the test so that it doesn't need `_.set`?  We'll also need to get approval for `cli-truncate` which should be fine.

shetzel (changes requested):

andreacorbellini:
> Hi @andreacorbellini we'll be happy to get this fixed. Can this be done with adding extra dependencies?

You mean with*out* adding extra dependencies? I think so, yes: I did some digging into `cli-truncate` and saw that it's a simple wrapper around `slice-ansi` with a few extra features that are not needed here. And `slice-ansi` is already in `yarn.lock`. Let's see if I can get it to work

andreacorbellini:
Done. The only new dependency now is `@types/slice-ansi`, and that's a dev-dependency that contains only annotations.

WillieRuemmele (approved):

WillieRuemmele:
Hey @andreacorbellini - this looks great 🏆 - we've got some testing PRs we're getting through, I'll wait for those to be merged and get this verified with the new tests and merge. Thank you so much for the contribution 😄

WillieRuemmele:
I verified this on my own branch with all new integration tests running ✅ 
https://github.com/oclif/core/pull/728

thank you for the contribution 🏆

PR #751, opened 2023-07-26
feat: deprecate scopedEnvVarKey for scopedEnvVarKeys which accounts for binAliases

WillieRuemmele:
@W-13825184@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -396,7 +396,7 @@ export class Config implements IConfig {
   }
 
   public scopedEnvVar(k: string): string | undefined {
-    return  process.env[this.scopedEnvVarKeys(k).find(k => process.env[k]) as string]
+    return process.env[this.scopedEnvVarKeys(k).find(k => process.env[k]) as string]
   }
 
   public scopedEnvVarTrue(k: string): boolean {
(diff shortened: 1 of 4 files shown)


test/config/config.test.ts
```diff
@@ -22,7 +22,7 @@ const pjson = {
   files: [],
   commands: {},
   oclif: {
-    binAliases: ['foo', 'bar'],
+    binAliases: ['bar', 'baz'],
```
WillieRuemmele, line 25: the bin is `foo`, changed to remove confusing assertions

src/config/config.ts
```diff
@@ -396,21 +396,36 @@ export class Config implements IConfig {
...
   }
 
   public scopedEnvVarTrue(k: string): boolean {
-    const v = process.env[this.scopedEnvVarKey(k)]
+    const v = process.env[this.scopedEnvVarKeys(k).find(k => process.env[k]) as string]
     return v === '1' || v === 'true'
   }
 
+  /**
+   * @deprecated - use scopedEnvVarKeys instead which will account for bin aliases, this could be private/removed in a major
```
mdonnalley, line 408: I'm not sure how I feel about deprecating this since the overwhelming majority of oclif CLIs won't ever use bin aliases. For them it would be annoying to have to deal with an array when it's always going to be an array with one element in it

I wonder if it might be better to keep it un-deprecated and just document clearly in the jsdoc that if they're using bin aliases they should use the other method

What do you think?

mdonnalley (approved):

PR #768, opened 2023-08-24
fix: use configured help class in --help

nigelzor:
without this change `my-cli command --help` always uses the default help class.


mdonnalley:
@nigelzor Thanks for the contribution!

PR #779, opened 2023-09-06
feat: final v3 features

### LINKED ISSUE #703: Improve flag alias handling
**Is your feature request related to a problem? Please describe.**
We can define flag aliases via the `aliases` prop in a flag definition.
This is expected to be an array of strings and can be either long or short flags, docs: 
https://oclif.io/docs/flags

As a side effect, oclif allows to use short aliases as the long name too:
Screenshot 2023-05-15 at 11 25 32

```
➜  plugin-org git:(main) ✗ sfdx force org create --u
Error (1): Flag --target-org expects a value
```

it also makes it harder to detect duplicate short/long flag names (have to compare flag names, short chars, long and short aliases):
https://github.com/oclif/plugin-command-snapshot/pull/328

**Describe the solution you'd like**
1. Allow to define short aliases in a new prop (`charAliases`)
2. Improve `flag.aliases` type to only allow strings longer than 1 char.
3 Restrict `charAliases` to strings.length === 1 only.

nice to have:
4. Improve flag types to make flag short/long name, aliases, etc collision a compile error.

### LINKED ISSUE #783: custom flag type doesn't respect `required` set in a custom flag
**Describe the bug**
If I have a custom flag with `required: true` and use it in a command, its type is `<T | undefined>`. 

**To Reproduce**

1. clone this branch of sf-plugin-api: https://github.com/cristiand391/sf-plugin-api/tree/cd%2Foclif-required-flag
2. see `target-org` type is `(property) 'target-org': OptionFlag<Org, CustomOptions>` [org api command](https://github.com/cristiand391/sf-plugin-api/blob/deb78ab0eccd81ba9043760d7d4fe9d36bb2b474/src/commands/org/api.ts#L36)
3. Delete this line: https://github.com/cristiand391/sf-plugin-api/blob/deb78ab0eccd81ba9043760d7d4fe9d36bb2b474/src/commands/org/api.ts#L39
4. `target-org` type now is: `(property) 'target-org': OptionFlag<Org | undefined, CustomOptions>`

`requiredOrgFlag` already sets `required: true` so we shouldn't need to set `required` again:
https://github.com/salesforcecli/sf-plugins-core/blob/c43456953f5fb14bb64d1cc72f84c78c2c211088/src/flags/orgFlags.ts#L154

For comparision, I added a two flags in the `org api` command:
https://github.com/cristiand391/sf-plugin-api/blob/deb78ab0eccd81ba9043760d7d4fe9d36bb2b474/src/commands/org/api.ts#L16-L23

`lalaFlag` is a string flag, `testFlag` is a custom flag and both set `required: true`, if you check their types here:
https://github.com/cristiand391/sf-plugin-api/blob/deb78ab0eccd81ba9043760d7d4fe9d36bb2b474/src/commands/org/api.ts#L31-L32

you can see `lalaFlag` respects `required:true`:
`(property) lala: OptionFlag<string, CustomOptions>`
and `testFlag` doesn't:
`(property) test: OptionFlag<string | undefined, CustomOptions>`

**Expected behavior**
custom flag type should recognize when `required` is set. 

 - OS & version: macos ventura
 - Shell/terminal & version zsh/alacritty

**Additional context**
We've been manually setting `required` in our plugins:
https://github.com/search?q=org%3Asalesforcecli+%2FFlags%5C.requiredOrg%2F+%2Frequired%3A+true%2F&type=code

You can see that `required` is actually passed (see below the help text org `--target-org` after `required: true` is deleted in `org api`), the issue seems to be only the flag type:

Screenshot 2023-09-12 at 12 36 07

mdonnalley:
- Adds `charAliases` flag prop (Fixes #703)
- Allows local `bin/dev.js` to auto-transpile ESM plugins (Fixes https://github.com/oclif/hello-world-esm/issues/38)
- Adds backwards compatibility for flexible taxonomy (some plugins may not have permutations cached in manifest yet)
- Caches command's `relativePath` and `isESM` properties in manifest
- Improve `perf` debug output (and make it more accuracte)
- Don't search for root of linked plugins (we already know what it is)
- Improved flag typing (defaults provided to `Flags.custom` are now respected Fixes #783 )
- Adds `Flags.option` which can infer the type from the provided options
```typescript
  static flags = {
    name: Flags.option({
      options: ['foo', 'bar'] as const,
    })(),
  }
}
```

@W-14082430@


shetzel (approved):

.github/workflows/test.yml
```diff
@@ -34,18 +32,56 @@ jobs:
...
+        with:
+          node-version: latest
+          cache: yarn
+      - run: yarn install --network-timeout 600000
+      - run: yarn build
+      - run: yarn link
+      - run: New-Item -Path D:\a -Name "e2e" -ItemType "directory"
+      - run: New-Item -Path D:\a\e2e -Name "sf.e2e.ts" -ItemType "directory"
+      - run: |
+          git clone https://github.com/salesforcecli/cli.git --branch mdonnalley/esm
```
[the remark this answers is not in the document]
mdonnalley: Yep. The sf integration tests fail since there are breaking changes. I have a note to switch this back to main once https://github.com/salesforcecli/cli/pull/1094 is merged

.github/workflows/test.yml
```diff
@@ -34,18 +32,56 @@ jobs:
...
+    strategy:
+      fail-fast: false
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: latest
+          cache: yarn
+      - run: yarn install --network-timeout 600000
```
mshanemc, line 52: here and elsewhere, prefer https://github.com/salesforcecli/github-workflows/blob/main/.github/actions/yarnInstallWithRetries/action.yml

use it like `      - uses: salesforcecli/github-workflows/.github/actions/yarnInstallWithRetries@main`

src/config/util.ts
```diff
@@ -14,22 +14,17 @@ export function mapValues<T extends Record<string, any>, TResult>(obj: {[P in ke
   }, {} as any)
 }
 
-export function exists(path: string): Promise<boolean> {
```
mshanemc, line 17: yay

src/config/util.ts
```diff
@@ -14,22 +14,17 @@ export function mapValues<T extends Record<string, any>, TResult>(obj: {[P in ke
...
-  // eslint-disable-next-line no-promise-executor-return
-  return new Promise(resolve => resolve(fs.existsSync(path)))
-}
-
 export function resolvePackage(id: string, paths: { paths: string[] }): string {
   return require.resolve(id, paths)
 }
 
-export function loadJSON(path: string): Promise<any> {
+export function loadJSON<T = unknown>(path: string): Promise<T> {
```
mshanemc, line 21: also yay for the <T>

src/config/util.ts
```diff
@@ -108,15 +103,15 @@ export function getCommandIdPermutations(commandId: string): string[] {
...
       const name = parts.join(':')
-      if (name) usuableIds.push(name)
+      if (name) usableIds.push(name)
       parts.pop()
     }
   }
 
-  return new Set(usuableIds)
+  return new Set(usableIds)
 }
```
mshanemc, line 117: immutable version

```suggestion
export const collectUsableIds = (commandIds: string[]): Set<string> =>
  new Set(commandIds.flatMap(id => commandIdToUsables(id)))

const commandIdToUsables = (command: string): string[] =>
  command.split(':').map((value, i, a) => a.slice(0, i + 1).join(':'))
```

src/errors/logger.ts
```diff
@@ -1,5 +1,5 @@
-import * as fs from 'fs/promises'
-import {dirname} from 'path'
+import {mkdir, appendFile} from 'node:fs/promises'
+import {dirname} from 'node:path'
 import stripAnsi = require('strip-ansi')
```
mshanemc, line 3: for some reason I though you'd need to get rid of `require` statements for ESM

src/interfaces/parser.ts
```diff
@@ -238,32 +242,118 @@ export type BooleanFlag<T> = FlagProps & BooleanFlagProps & {
...
-  default?: FlagDefault<T[] | undefined, P>;
-  multiple: true;
-})
+}
 
-export type FlagDefinition<T, P = CustomOptions> = {
+type ReturnTypeSwitches = {multiple: boolean; requiredOrDefaulted: boolean}
+
+/**
+ * The logic here is as follows:
```
mshanemc, line 255: that's really clear and nice

src/command.ts
```diff
@@ -315,7 +315,7 @@ export abstract class Command {
       }
 
       const deprecateAliases = flagDef?.deprecateAliases
-      const aliases = (flagDef?.aliases ?? []).map(a => a.length === 1 ? `-${a}` : `--${a}`)
+      const aliases = ([...flagDef?.aliases ?? [], ...flagDef?.charAliases ?? []]).map(a => a.length === 1 ? `-${a}` : `--${a}`)
```
mshanemc, line 318: is it possible to put, for example, `c` in both aliases and charAliases such that they'd get the warning multiple times (and we should Set-dedupe them?)

also, rather than iterating all these aliases, should we exit early if `!deprecateAliases` and then compute the aliases and check length?

src/flags.ts
```diff
@@ -121,3 +151,70 @@ export const help = (opts: Partial<BooleanFlag<boolean>> = {}): BooleanFlag<void
     },
   })
 }
+
+type ElementType<T extends ArrayLike<unknown>> = T[number];
```
mshanemc, line 155: what's the logic for `ArrayLike` vs `Array` ?  It's an es5 polyfill

mdonnalley: @mshanemc It's an artifact of a StackOverflow copy/paste :)

I'll update it to `ReadonlyArray`, which is really what we want here

src/main.ts
```diff
@@ -1,9 +1,9 @@
-import {fileURLToPath} from 'url'
+import {fileURLToPath} from 'node:url'
 
-import {format, inspect} from 'util'
+import {format, inspect} from 'node:util'
 
 import * as Interfaces from './interfaces'
-import {URL} from 'url'
+import {URL} from 'node:url'
```
mshanemc, line 6: maybe organize these, put all the `node:` bits first

src/module-loader.ts
```diff
@@ -79,7 +80,32 @@ export default class ModuleLoader {
     let isESM: boolean | undefined
     try {
       ({isESM, filePath} = ModuleLoader.resolvePath(config, modulePath))
-      const module = isESM ? await import(url.pathToFileURL(filePath).href) : require(filePath)
+      const module = isESM ? await import(pathToFileURL(filePath).href) : require(filePath)
```
mshanemc, line 83: meta: this whole static-only class bothers me.  You could have a 2 exported functions (`load``, `loadWithData`, `loadWithDataFromManifest`) and then some non-exported ones.

By breaking these up, you'd have the opporunity to only `require/import` what's needed instead of doing the whole "class" 

And this isn't a top-level export, but that would prevent poeple from using functions like `resolvePath` outside of this.

test/interfaces/flags.test-types.ts
```diff
@@ -4,7 +4,7 @@
 
 import {Command, Flags, Interfaces} from '../../src'
 import {expectType, expectNotType} from 'tsd'
-import {URL} from 'url'
+import {URL} from 'node:url'
```
mshanemc, line 7: your type-testing thing just keeps paying dividends!

test/interfaces/flags.test-types.ts
```diff
@@ -45,85 +100,242 @@ class MyCommand extends BaseCommand {
...
+    })({
+      multiple: true,
+      required: true,
+    }),
+
+    'custom#defs:multiple=true;opts:multiple=false,default': Flags.custom({
+      multiple: true,
+    })({
+      multiple: false,
+      // TODO: THIS IS A BUG. It should enforce a single value instead of allowing a single value or an array
```
mshanemc, line 336: did you mean to fix it for this major version?

test/parser/parse.test.ts
```diff
@@ -1748,5 +1748,16 @@ See more help with --help`)
       })
       expect(out.flags.foo).to.equal(true)
     })
+
+    it('works with aliased short char', async () => {
+      const out = await parse(['-b'], {
+        flags: {
+          foo: Flags.boolean({
+            charAliases: ['b'],
```
mshanemc, line 1756: UT coverage ideas:
do a non-boolean flag
empty array
same char duplicated in aliases and charAliases
error: using the full flag and its charAliases

src/errors/logger.ts
mshanemc: ...so v3 enables ESM for clis/plugins but isn't itself ESM, right?  When does that need to happen?

test/parser/parse.test.ts
mshanemc: I'll add these in a PR

test/interfaces/flags.test-types.ts
mdonnalley: @mshanemc I'm probably not going to fix it in this PR (if at all). It's a very odd use case... someone defines a custom flag with `multiple: true` and then explicitly sets `multiple: false` when using the flag AND sets a default. Given my experience updating the flag types so far, I'd say it's likely a 2 day effort to fix this one edge case

src/errors/logger.ts
mdonnalley: @mshanemc Correct. V3's major feature is that it fully supports CJS + ESM interoperability. The package itself is not ESM though.

Keeping it CJS for now allows CJS plugins to use v3, which makes the migration path for us and others much easier. I'm open to reconsidering that position though.

PR #785, opened 2023-09-13
build(ejs): bump ejs v to 3.1.9

filiar:
Bump ejs from v3.1.8 to 3.1.9 as BlackDuck scan reports security vulnerabilities for the v3.1.8 and recommends using 3.1.9


filiar:
CLA should be fixed

PR #797, opened 2023-09-27
feat!: collect perf results from outside oclif

mshanemc:
Allows plugins/libraries to consume Performance.

BREAKING CHANGE: Performance.mark(),  Perforance.getResult() now requires an `owner` property meant to be the npm package name.

has a special getter for `oclifPerf` 
(OclifHighlights, which does most of the transformations formerly in plugin-telemetry)

---

QA:  this is painful, sorry.

check out this branch and `yarn link`
in your local CLI directly, `yarn link @oclif/core`

You should be able to run commands with `DEBUG=oclif-perf` and see the oclif output.  There's nothing going to AppInsights OR any non-oclif results.

if you want to validate what goes into AppInsights,
yarn link this branch into plugin-telemetry using this PR https://github.com/salesforcecli/plugin-telemetry/pull/515
sf plugins link [plugin-telemetry]
then run some command with `SF_TELEMETRY_DEBUG=true DEBUG=sfdx:telemetry*` so you can get your CLIId to use for searching appInsights logs

Oh, you actually want to try the feature (peformance marks outside of oclif) ??
It's easiest to put these in a linked plugin (ex: PDR).  You have to use plugin linked into sf so that you get all the hooks (`bin/dev` won't do)
I have some already created for SDR in https://github.com/forcedotcom/source-tracking/pull/478, but you'd need to yarn link oclif/core into it and then yarn link that into your linked plugin

@W-14182563@


src/performance.ts
```diff
@@ -12,17 +12,21 @@ type PerfResult = {
 }
 
 type PerfHighlights = {
-  configLoadTime: number;
```
mshanemc, line 15: these will now be pre-formatted for AppInsights, which doesn't like nested obj.

src/performance.ts
```diff
@@ -12,17 +12,21 @@ type PerfResult = {
...
+  'oclif.initMs': number;
+  'oclif.commandLoadMs': number;
+  'oclif.initHookMs': number;
+  'oclif.prerunHookMs': number;
+  'oclif.postrunHookMs': number;
+  'oclif.commandRunMs': number;
+  'oclif.corePluginsLoadMs': number;
+  'oclif.userPluginsLoadMs': number;
+  'oclif.linkedPluginsLoadMs': number;
   pluginLoadTimes: Record<string, {duration: number, details: Details}>;
```
mshanemc, line 26: telemetry.record will skip object types and not record them to the telemetry file

src/performance.ts
```diff
@@ -202,33 +215,46 @@ export class Performance {
   public static debug(): void {
     if (!Performance.enabled) return
 
-    const debug = require('debug')('perf')
-    debug('Total Time: %sms', Performance.highlights.runTime.toFixed(4))
```
mshanemc, line 206: this was creating highlights and then letting plugin-telemetry re-massage the data.  Now we only do it once.

src/performance.ts
```diff
@@ -202,33 +215,46 @@ export class Performance {
...
+        oclifDebug(`    ${plugin}: ${duration.toFixed(4)}ms`)
       }
     }
 
-    debug('Command Load Time: %sms', Performance.highlights.commandLoadTime.toFixed(4))
-    debug('Command Run Time: %sms', Performance.highlights.commandRunTime.toFixed(4))
+    oclifDebug('Command Load Time: %sms', Performance.oclifPerf['oclif.commandLoadMs'].toFixed(4))
+    oclifDebug('Command Run Time: %sms', Performance.oclifPerf['oclif.commandRunMs'].toFixed(4))
+
+    const nonOclifDebug = require('debug')('non-oclif-perf')
```
mshanemc, line 247: separate debug makes it easier to only get the perf stuff you want.

src/config/config.ts
```diff
@@ -9,9 +9,9 @@ import {URL, fileURLToPath} from 'node:url'
 import {arch, userInfo as osUserInfo, release, tmpdir, type} from 'node:os'
 import {compact, ensureArgObject, getHomeDir, getPlatform, isProd, requireJson} from '../util'
 import {join, sep} from 'node:path'
-
+// eslint-disable-next-line sort-imports
```
mdonnalley, line 12: Why disable?

src/performance.ts
mdonnalley, line 247: This is a misleading name. What if we end up putting a bunch of markers in `@oclif/plugin-update`?

src/performance.ts
```diff
@@ -202,33 +215,46 @@ export class Performance {
...
+
+    const nonOclifPerf = Performance.results
+    if (nonOclifPerf.size > 0) {
+      nonOclifDebug('Non-Oclif Performance Measurements:')
```
mdonnalley, line 251: I think it'd be more accurate to say `Non-Core Performance Measurements:`

src/config/config.ts
mshanemc: Try it without.  There's nothing that make it happy

src/performance.ts
mshanemc: Yeah, I'll rename that other one, too.

src/config/config.ts
mshanemc: ok, now it works happily.  Cool, nevermind.

mdonnalley (approved):

PR #798, opened 2023-09-28
chore: use prettier and lint-staged

mdonnalley:
- Use prettier to make things pretty
- Use lint-staged to make sure no new changes aren't pretty
- Reformat all files so that they're pretty
- Move ansi-escapes require to top level of spinner.ts (otherwise it gets required every 100ms when the spinner is re-rendered)


mshanemc (approved):

.eslintrc.json
```diff
@@ -1,7 +1,11 @@
 {
   "extends": [
     "oclif",
-    "oclif-typescript"
+    "oclif-typescript",
+    "prettier"
+  ],
```
mshanemc, line 6: see prettier's recommendations for how to set it up.  The eslint **plugin** is not recommended.  We don't do it that way in salesforcecli, either.

https://prettier.io/docs/en/integrating-with-linters.html

PR #811, opened 2023-10-05
fix: only Command.Loadable in Help

mdonnalley:
- Update `execute` examples
- Only use `Command.Loadable` in `Help` code
- Un-fancy-ify `formatCommand` tests

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -378,15 +378,30 @@ export abstract class Command {
 }
 
 export namespace Command {
+  /**
+   * The Command class exported by a command file.
+   */
   export type Class = typeof Command & {
     id: string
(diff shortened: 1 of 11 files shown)


mshanemc (approved):

src/command.ts
```diff
@@ -383,7 +383,7 @@ export namespace Command {
     run(argv?: string[], config?: LoadOptions): Promise<any>
   }
 
-  export interface Loadable extends Cached {
+  export type Loadable = Cached & {
```
mshanemc, line 386: I'd like a jsdoc explanation of what this type is meant to be used for, since lots of function params refer to it.

PR #813, opened 2023-10-09
fix: custom flags with union type

### LINKED ISSUE #812: Custom flags with multiple: false
Do you want to request a *feature* or report a *bug*?
-----------------------------------------------------
Bug, I think.

What is the current behavior?
-----------------------------
Consider the following custom Flag:

```ts
export const versionSchemeFlag = Flags.custom<VersionScheme | undefined>({
	description: "Version scheme to use.",
	options: ["semver", "internal", "virtualPatch"],
	parse: async (input) => {
    	if (isVersionScheme(input)) {
			return input;
		}
	},
});
```

This declaration does not compile with oclif 4.0.0. The compiler complains that `multiple` is missing from the declaration, and `multiple: false` is not permitted. I have to set `multiple: true`. In other words, this DOES compile:

```ts
export const versionSchemeFlag = Flags.custom<VersionScheme | undefined>({
	description: "Version scheme to use.",
	options: ["semver", "internal", "virtualPatch"],
	parse: async (input) => {
    	if (isVersionScheme(input)) {
			return input;
		}
	},
	multiple: true,
});
```

What is the expected behavior?

In prior versions it was possible to declare a custom flag with multiple=false. I would expect to be able to do this in the most recent version.

oclif version
------------

oclif/4.0.0 linux-x64 node-v18.18.0

mdonnalley:
Fixes issue with custom flags with union type unnecessarily enforcing `multiple: true`

Fixes #812 

@W-14261473@

### DIFF
--- a/src/flags.ts
+++ b/src/flags.ts
@@ -145,30 +145,30 @@ export const help = (opts: Partial<BooleanFlag<boolean>> = {}): BooleanFlag<void
     },
   })
 
-type ElementType<T extends ReadonlyArray<unknown>> = T[number]
+type ReadonlyElementOf<T extends ReadonlyArray<unknown>> = T[number]
 
 export function option<T extends readonly string[], P extends CustomOptions>(
-  defaults: Partial<OptionFlag<ElementType<T>[], P>> & {
+  defaults: Partial<OptionFlag<ReadonlyElementOf<T>[], P>> & {
     multiple: true
     options: T
   } & (
       | {
-          default: OptionFlag<ElementType<T>[], P>['default'] | undefined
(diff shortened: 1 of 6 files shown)


src/flags.ts
```diff
@@ -145,39 +145,39 @@ export const help = (opts: Partial<BooleanFlag<boolean>> = {}): BooleanFlag<void
     },
   })
 
-type ElementType<T extends ReadonlyArray<unknown>> = T[number]
+type ReadonlyElementOf<T extends ReadonlyArray<unknown>> = T[number]
```
mdonnalley, line 148: This isn't part of the fix - I just thought it would be a clearer name for the type

shetzel (dismissed):

Issue #817, opened 2023-09-09
./bin/dev.js throws `ERR_UNKNOWN_FILE_EXTENSION` on Node 20


bader-nasser:
After cloning the project and installing the deps using `pnpm`, running `node ./bin/dev.js help` does not show the commands!

Also, when I change any thing in the commands (e.g. description) and run `node ./bin/dev.js help hello`, I don't get the new values.

mdonnalley:
@bader-nasser I transferred this issue to `@oclif/core` since this issue will be more visible to others here. 

It seems that node 20 introduced a breaking change to the ESM loader hooks, which affects `ts-node`'s ability to transpile ESM code.

>Custom ESM loader hooks run on dedicated thread
ESM hooks supplied via loaders (--experimental-loader=foo.mjs) now run in a dedicated thread, isolated from the main thread. This provides a separate scope for loaders and ensures no cross-contamination between loaders and application code.

[Source](https://nodejs.org/en/blog/announcements/v20-release-announce#custom-esm-loader-hooks-nearing-stable)

ts-node issue: https://github.com/TypeStrong/ts-node/issues/1997

Obviously, there's not much we can do on the oclif side to fix this. The best we'll be able to do is provide a better error message and provide a couple workarounds:

**Workarounds**
- Downgrade to Node 19 or older
- Replace `ts-node` in `bin/dev.js` with `#!/usr/bin/env -S node --loader ts-node/esm --no-warnings=ExperimentalWarning`

bader-nasser:
great. I'm glad that the root issue is identified.

PR #818, opened 2023-10-10
fix: warn about ts-node and node 20

mdonnalley:
- Warn when `ts-node` is being used to execute `bin/dev.js` and node is >= 20
- Fall back to compiled source when `ts-node` cannot be used

Fixes https://github.com/oclif/core/issues/817

@W-14270164@

### DIFF
--- a/src/config/ts-node.ts
+++ b/src/config/ts-node.ts
@@ -1,4 +1,3 @@
-/* eslint-disable complexity */
 import {existsSync} from 'node:fs'
 import {join, relative as pathRelative, sep} from 'node:path'
 import * as TSNode from 'ts-node'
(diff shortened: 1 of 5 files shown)


src/config/ts-node.ts
```diff
@@ -1,5 +1,6 @@
+/* eslint-disable complexity */
```
WillieRuemmele, line 1: I don't think this is required - it passed linting for me locally

src/execute.ts
```diff
@@ -34,7 +34,7 @@ import {settings} from './settings'
  *
  * @example For CJS dev.js
  * ```
- * #!/usr/bin/env ts-node
+ * #!/usr/bin/env -S node --loader ts-node/esm --no-warnings=ExperimentalWarning
```
WillieRuemmele, line 37: I'll defer to you, but all of this ESM stuff is new to me... but I would've expected the CJS to remain the same?

WillieRuemmele (approved):

PR #830, opened 2023-10-15
Cache env

amcaplan:
Background: In the [Shopify CLI](https://github.com/Shopify/cli), we allow specifying every flag as an environment variable. We want to include that information in our help output. However, since we (correctly) use manifests in production, `env` isn't available for display because it isn't included in the cached flags.

I don't see any reason specifically not to cache the flags' `env` values.

What do you think?


mdonnalley (approved):

mdonnalley:
@amcaplan This great! Thank you for the PR

PR #832, opened 2023-10-16
feat: no longer use wrapped process.stdout and process.stderr

### LINKED ISSUE #799: ux.action.spinner overwrites most recent terminal output when updating spinner
**Describe the bug**

The last line outputted to the terminal is overwritten when the spinner updates. 

I've tracked it down to this line of code, https://github.com/oclif/core/blob/2.15.0/src/cli-ux/action/spinner.ts#L78, which uses ansi-escapes to move the cursor up and then erase everything underneath it. The underlying assumption is that it's moving the cursor to the beginning of the existing spinner, but that assumption fails once additional lines have been made to the terminal.

One potential solution is to use `ansiEscapes.cursorSavePosition` to capture the position of the cursor immediately after rendering the spinner and then use `ansiEscapes.cursorRestorePosition` to restore that position just before re-rendering it. I tested this and it did work until the spinner completed, at which point it cleared the additional logs. Maybe there's something to go off there.

**To Reproduce**

```typescript
console.log('hello world 1') // this stays on the terminal
ux.action.start('starting spinner')
console.log('hello world 2')  // this stays on the terminal
console.log('hello world 3') // this gets overwritten once spinner updates
await ux.wait(2500)
ux.action.status = 'still going'
await ux.wait(2500)
ux.action.stop()
```

```
hello world 1
starting spinner...
hello world 2
starting spinner... done
```

mdonnalley:
When a spinner is started, it creates a mock version of `process.stdout.write` and `process.stderr.write` to capture _all_ log messages (from `console.log`, `debug`, `ux.log`, etc...) and then flushes the captured logs just before re-rendering the spinner.

By encapsulating `process.stdout` and `process.stderr` (introduced [here](https://github.com/oclif/core/pull/629)), the spinner was no longer able to capture all logs messages - just the ones that went through the encapsulated `process.stdout` and `process.stderr`. This means that anything sent to the terminal via `console.log` or `debug` was at risk of being overwritten when the spinner re-render. 

The solution here is to back out of using the encapsulated `process.stdout` and `process.stderr`. These were originally added so that it would be easier to stub `process.stdout|stderr.write` in unit tests (see [mocha issue](https://github.com/mochajs/mocha/issues/1582) related to stubbing `process.stdout.write`). In place of that, I've reduced the usage of `process.stdout.write` to `write.stdout` and `write.stderr` and added `makeStubs` method to `ux` that people can use if they need to stub those.

I've also deprecated the `Stream` class that was used to encapsulate `process.stdout` and `process.stderr` and will be removed in the next major version

Fixes #799 
Closes #828 
(Passing tests depend on https://github.com/oclif/core/pull/836)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,31 @@
+## [3.3.3](https://github.com/oclif/core/compare/3.3.2...3.3.3) (2023-10-19)
+
+
+### Bug Fixes
+
+* allow github-installed plugins to be auto-transpiled ([6f2c5e2](https://github.com/oclif/core/commit/6f2c5e2ec9b9bafa7dea4ae2e7350ada64ff0e88))
+* handle ModuleLoadError from hooks ([0321093](https://github.com/oclif/core/commit/0321093e4f8a0f659346a9e2e808ad44c6d3771f))
+
+
+
+## [3.3.2](https://github.com/oclif/core/compare/3.3.1...3.3.2) (2023-10-17)
+
+
+### Bug Fixes
+
+* **deps:** bump @babel/traverse from 7.23.0 to 7.23.2 ([d53498a](https://github.com/oclif/core/commit/d53498af1da6bdcb3fb70f7c0b6e0da8da0831a1))
+
+
+
+## [3.3.1](https://github.com/oclif/core/compare/3.3.0...3.3.1) (2023-10-16)
+
+
+### Bug Fixes
+
+* add frequency and frequencyUnit to warn-if-update ([#827](https://github.com/oclif/core/issues/827)) ([3567c74](https://github.com/oclif/core/commit/3567c748a8a04ab94cca493aa1dfcb4f10370f6e))
+
+
+
 # [3.3.0](https://github.com/oclif/core/compare/3.2.1...3.3.0) (2023-10-16)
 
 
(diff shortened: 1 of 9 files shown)


WillieRuemmele (dismissed):

WillieRuemmele:
### QA Notes
---
linked this into the demo plugin used in the issues repo:

before
```bash
➜  plugin-demo git:(main) ✗ 
 ➜  ./bin/dev demo
hello world 1
spinner... ⣻
hello world 2
spinner... done
```

after
✅ 
```bash
➜  plugin-demo git:(main) ✗ 🔗1
 ➜  ./bin/dev demo                                                                    
hello world 1
hello world 2
hello world 3
spinner... done
```

linked into `plugin-source` doing a `force source pull` with a remote change

before
```bash
 ➜  sf force source pull         
Warning: We plan to deprecate this command in the future. Try using the "project retrieve start" command instead.
Loading source tracking information... ⣟
Loading source tracking information... ⣽ Retrieving metadata from the org
Loading source tracking information... ⡿ Checking for deletes from the org and updating source tracking files
Updating source tracking...
Updating source tracking... done
```

after
✅ 
```bash
 ➜  sf force source pull
Warning: We plan to deprecate this command in the future. Try using the "project retrieve start" command instead.
Loading source tracking information... ⣷
Updating source tracking... done
```

WillieRuemmele (approved):

PR #845, opened 2023-10-27
feat: read extended tsconfigs

mdonnalley:
- Pass entirety of `tsconfig` to `tsNode.register`
- Use `tsconfck` instead of `typescript` for reading/merging extended tsconfigs


src/module-loader.ts
```diff
@@ -38,7 +38,7 @@ export async function load<T = any>(config: IConfig | IPlugin, modulePath: strin
   let filePath: string | undefined
   let isESM: boolean | undefined
   try {
-    ;({filePath, isESM} = resolvePath(config, modulePath))
+    ;({filePath, isESM} = await resolvePath(config, modulePath))
```
mshanemc, line 41: what's the semicolon for?

test/integration/esm-cjs.ts
```diff
@@ -406,43 +409,43 @@ type PluginConfig = {
   }
 
   const cjsTests = async () => {
-    await test('Install CJS plugin to CJS root plugin', async () => {
-      await installTest(PLUGINS.cjs2, cjsExecutor)
-    })
+    // await test('Install CJS plugin to CJS root plugin', async () => {
```
mshanemc, line 412: why are these commented out?

src/config/plugin.ts
```diff
@@ -41,15 +41,34 @@ const search = (cmd: any) => {
...
+}
+
 export class Plugin implements IPlugin {
   alias!: string
 
   alreadyLoaded = false
 
   children: Plugin[] = []
 
+  commandIDs!: string[]
```
mshanemc, line 66: doesn't the code below break the interface?  When it was a getter, it would always return something.  The `!` means that it's not possible to have a Plugin (class) without commandIDs, but without calling `load()` you would.

Maybe init them as empty arrays and then load can fix them if it runs?

src/config/plugin.ts
```diff
@@ -162,8 +150,25 @@ export class Plugin implements IPlugin {
     return cmd
   }
 
+  public async getCommandIDs(): Promise<string[]> {
```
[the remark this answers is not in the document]
mshanemc: they're not in the interface so maybe `private` or `#getCommandIDs` is appropriate?

or a function that takes the `name` and `commandsDir` ?

src/config/ts-node.ts
```diff
@@ -15,42 +15,45 @@ const debug = Debug('ts-node')
...
+
+    if (tsconfig.extends) {
+      const {parse} = await import('tsconfck')
+      const result = await parse(tsconfigPath, {})
+      let tsNodeOpts = {}
+      for (const extended of (result.extended ?? []).reverse()) {
+        if (extended.tsconfig['ts-node']) {
+          tsNodeOpts = {...tsNodeOpts, ...extended.tsconfig['ts-node']}
+        }
+      }
```
mshanemc, line 36: not 100% sure this works, but fromEntries() is LWW so you don't have to reverse the array to get the last ones in first for Object.assign.

```ts
const tsNodeOpts = Object.fromEntries(
        (result.extended ?? [])
          .map((e) => Object.entries(e.tsconfig['ts-node'] ?? {}))
      )
```

mshanemc (changes requested):
requested changes are the public api questions and interface stuff

test/integration/esm-cjs.ts
mdonnalley: Commented out for iterating on tests - thanks for catching!

src/module-loader.ts
mdonnalley: It's a default prettier thing

src/config/plugin.ts
```diff
@@ -192,7 +180,14 @@ export class Plugin implements IPlugin {
...
     }
 
-    this.hooks = mapValues(this.pjson.oclif.hooks ?? {}, (i) => castArray(i).map((i) => tsPath(this.root, i, this)))
+    this.commandsDir = await this.getCommandsDir()
+    this.commandIDs = await this.getCommandIDs()
+
+    this.hooks = {}
+    for (const [k, v] of Object.entries(this.pjson.oclif.hooks ?? {})) {
+      // eslint-disable-next-line no-await-in-loop
+      this.hooks[k] = await Promise.all(castArray(v).map(async (i) => tsPath(this.root, i, this)))
```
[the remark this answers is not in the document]
mshanemc:
```ts
this.hooks = Object.fromEntries(await Promise.all(
      Object.entries(this.pjson.oclif.hooks ?? {})
        .map(async ([k, v]) => [k, await Promise.all(
          castArray(v).map(async (i) => tsPath(this.root, i, this))
        )])
    ))
```

mshanemc (dismissed):

mshanemc:
QA notes:

linked this PR into plugin-user to use bin/dev (cjs), 

`./bin/dev org:list:users -o hub`
✅ works

set `./tsconfig.json` to have
```json
{ 
  "ts-node": {
    "transpileOnly": true
  }
}
```

`DEBUG=* ./bin/dev org:list:users -o hub` to check out the debug logs
✅ I see   `config:ts-node   transpileOnly: true,`

---
in main/top-level tsconfig, showConfig:true. In what it extends (`dev-config/tsconfig-strict.json`) it's false

`DEBUG=* ./bin/dev org:list:users -o hub` to check out the debug logs
❌  I see   `config:ts-node   showConfig: false,`

I think I was expecting true--like the top-level one is extends other ones, but would override them.

mshanemc (approved):

mshanemc:
QA round 2:

linked into plugin-user.

All permutations of top-level tsconfig, plus local modifications to dev-config/tsconfig-strict which extends dev-config/tsconfig.

✅ all working with proper results when toggling a single ts-node prop (showConfig)
✅ can add other props in "lower" levels and have them preserved when higher levels don't include that one.

PR #852, opened 2023-11-05
Enable themes

AllanOricil:
With this feature developers will be able to create custom color themes for their CLI.

obs: disregard the `--json` I added in the pictures below

### BEFORE

image

image

image

image

### AFTER

image

image

image

image

-----

### A theme can be enabled when:
- an env variable called `<BIN>_ENABLE_THEME` is set
- `enableTheme` property is set on the package.json

````json
{
   ...
   "oclif": {
       "enableTheme": true
   }
   ...
}
````

image

### All theme colors are optional
image

----

## Configurable Tokens

Colors are set with the values from `~/.config/<CLI>/theme.json` file, if one is present. Colors can be set with Hex and RGB color codes strings, which are converted to colors using the [color package](https://www.npmjs.com/package/color).

````json
{
       "alias" : "red",
       "bin" : "rgb(255, 255, 255)",
       "command" : "#45C65A",
       "commandSummary" : "#FFFFFF",
       "dollarSign" : "#FFFF00",
       "flag" : "#45C65A",
       "flagDefaultValue" : "#1AB9FF",
       "flagOptions" : "#45C65A",
       "flagRequired" : "#FE5C4C",
       "flagSeparator" : "#FFFFFF",
       "flagType" : "#0EAEE8",
       "sectionDescription" : "#FFFFFF",
       "sectionHeader" : "#FFFF00",
       "topic" : "#45C65A",
       "version" : "#45C65A"
}
````

In the following pictures you can understand the regions each token affect.

`alias`
image

`bin`
image

`command`
image

image

`commandSummary`

image

image

`dollarSign`
image

`flag`
image

`flagDefaultValue`
image

`flagOptions`
image

`flagRequired`
image

`flagSeparator`
image

`flagType`  

[NOT IMPLEMENTED YET. ](https://www.youtube.com/watch?v=dQw4w9WgXcQ)

It depends on a new attribute called "typeLabel" which has to be added to the flag configuration. Without allowing CLI developers to specify a custom label for the types of their flags, like `filePath`, users would only see "boolean | option" which is the value from `flag.type`.

`sectionDescription`
image

`sectionHeader`
image

`topic`
image

`version`
image

### CODE REVIEW

**Comment by AllanOricil** [comment]
Some tests are failing but I'm ready for a review since the feature is mostly done.

**Comment by AllanOricil** [comment]
I found that the `DESCRIPTION`, `ALIASES`, and `EXAMPLES` sections aren't working that well with the Heroku cli. I'm going to fix it, now that I have means to test it. But I could not find a command which would display `FLAG DESCRIPTIONS` and `ARGUMENTS`. It would be good if you could help me to find a CLI that has these sections.

image

File: src/config/config.ts [hunk 1]
```diff
@@ -312,6 +340,12 @@ export class Config implements IConfig {
     if (this.pjson.oclif.topicSeparator && [' ', ':'].includes(this.pjson.oclif.topicSeparator))
       this.topicSeparator = this.pjson.oclif.topicSeparator!
     if (this.platform === 'win32') this.dirname = this.dirname.replace('/', '\\')
+
+    this.enableTheme = process.env.OCLIF_ENABLE_THEME === 'true' || this.pjson.oclif.enableTheme
```
mdonnalley, line 344: this should be 

```typescript
this.enableTheme = this.scopedEnvVarTrue('ENABLE_THEME') ?? this.pjson.oclif.enableTheme
```

mdonnalley (changes requested):
@AllanOricil Great work 🏆 

My biggest concern here is that users will have different terminal themes that would conflict with the configured theme. CLI developers would need to configure a theme that works on most terminal themes, which would be a challenging task.

I wonder if this should be something that we defer to users? Instead of looking in the CLI's package.json, oclif/core could look for a config file on the users machine. That would allow users to have colorful help without forcing them to use a specific terminal theme that works with it.

What do you think?

AllanOricil:
> @AllanOricil Great work 🏆
> 
> My biggest concern here is that users will have different terminal themes that would conflict with the configured theme. CLI developers would need to configure a theme that works on most terminal themes, which would be a challenging task.
> 
> I wonder if this should be something that we defer to users? Instead of looking in the CLI's package.json, oclif/core could look for a config file on the users machine. That would allow users to have colorful help without forcing them to use a specific terminal theme that works with it.
> 
> What do you think?

Makes sense. Do you have an example of how you usually read files from the User's computer?
I was also thinking about installing a package to set the theme and then use an env variable `THEME_NAME` to point to the package that has the theme. What do you think?

AllanOricil:
> @AllanOricil Great work 🏆
> 
> My biggest concern here is that users will have different terminal themes that would conflict with the configured theme. CLI developers would need to configure a theme that works on most terminal themes, which would be a challenging task.
> 
> I wonder if this should be something that we defer to users? Instead of looking in the CLI's package.json, oclif/core could look for a config file on the users machine. That would allow users to have colorful help without forcing them to use a specific terminal theme that works with it.
> 
> What do you think?

The `theme` prop is for CLI developers to ship a default theme. The package that has the theme will be used to overwrite the default colors set by the CLI owner. For example, `https://github.com/salesforcecli/cli/blob/main/package.json` will ship with a default `theme` and disabled. Then Users can enable it using `OCLIF_ENABLE_THEME`, and they can also overwrite the default theme installing a package and using the `OCLIF_THEME_NAME` variable to set the theme.

mdonnalley:
@AllanOricil you can use the [`safeReadJson`](https://github.com/oclif/core/blob/main/src/util/fs.ts#L65) function to read the file. I think the config file would live in `this.config.configDir` which typically resolves to `~/.config/<CLI>`

I don't think setting a default theme is a smart idea - even if it's not enabled unless the user sets the env var. I'm imagining a scenario where a user sets that env var for the first time and then suddenly half the help text isn't visible because it's white text on a white background. They then have to figure out what went wrong or they assume that the CLI is broken. I think that's a bad user experience.

A better user experience, in my opinion, is for CLIs to document for their users how to set a theme for their help text and then offer a few variations (e.g. a dark theme, a light theme, etc.)

If the CLI developer wanted to have that more automatic for users then they could create a plugin that allows the user to automatically set their theme to one of their recommended themes, e.g. `<CLI> theme set --theme=dark`. This could even be an oclif plugin - but that's outside the scope of this PR

AllanOricil:
> @AllanOricil you can use the [`safeReadJson`](https://github.com/oclif/core/blob/main/src/util/fs.ts#L65) function to read the file. I think the config file would live in `config.configDir` which typically resolves to `~/.config/<CLI>`
> 
> I don't think setting a default theme is a smart idea - even if it's not enabled unless the user sets the env var. I'm imagining a scenario where a user sets that env var for the first time and then suddenly half the help text isn't visible because it's white text on a white background. They then have to figure out what went wrong or they assume that the CLI is broken. I think that's a bad user experience.
> 
> A better user experience, in my opinion, is for CLIs to document for their users how to set a theme for their help text and then offer a few variations (e.g. a dark theme, a light theme, etc.)
> 
> If the CLI developer wanted to have that be more automatic for their users then they could create a plugin that allows the user to automatically set the theme to one of their recommended themes, e.g. `<CLI> theme set --theme=dark`. This could even be an oclif plugin - but that's outside the scope of this PR

Yes, you are right. I will do what you suggested because it is much better. Thanks

AllanOricil:
@mdonnalley 
I can't really test the `FLAG DESCRIPTIONS ` section because I don't know any command that has flags without both summary and description. Do you know any? 

image

AllanOricil:
I added colors to `ALIAS`, `ARGUMENTS`, `DESCRIPTION`, `EXAMPLES`, and `FLAG DESCRIPTIONS`.

<b>obs:</b> I can't really do much for the `EXAMPLE` section because I can't separate the "command" from the whole example. So I decided to only tokenize "$" to help people to at least differentiate between lines
<b>obs:</b> I can't test the `FLAG DESCRIPTIONS` section because I don't know any command that displays it. Its description is set using the `sectionDescription` color.

image

image

src/config/config.ts
AllanOricil: fixed

AllanOricil:
All tests from `main` are working again. If everything is OK, I will start writing tests for my changes.

image

mdonnalley:
>I can't really test the FLAG DESCRIPTIONS  section because I don't know any command that has flags without both summary and description. Do you know any?

`sf project deploy start` has a `FLAG DESCRIPTIONS` section

mdonnalley (changes requested):

mdonnalley:
@AllanOricil I think the failing `esm-cjs` test is because of one of your changes:

```
  e2e:esm-cjs.ts:cjs:plugin-test-cjs-1 stdout $ shx rm -rf dist && tsc -b
node_modules/@oclif/core/lib/interfaces/config.d.ts(2,8): error TS1259: Module '"/private/var/folders/4l/6tqf4pm156j751spjj1qppwr0000gp/T/esm-cjs.ts/cjs/plugin-test-cjs-1/node_modules/@oclif/core/node_modules/@types/color/index"' can only be default-imported using the 'esModuleInterop' flag
node_modules/@oclif/core/lib/help/util.d.ts(1,8): error TS1259: Module '"/private/var/folders/4l/6tqf4pm156j751spjj1qppwr0000gp/T/esm-cjs.ts/cjs/plugin-test-cjs-1/node_modules/@oclif/core/node_modules/@types/color/index"' can only be default-imported using the 'esModuleInterop' flag
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

That seems to be saying that the `colors` package cannot be default imported. Probably just needs to be changed to `import * as Color from 'color'`

yarn.lock
```diff
@@ -1012,11 +1012,25 @@
   dependencies:
     "@types/node" "*"
 
+"@types/color-convert@*":
+  version "2.0.2"
+  resolved "http://localhost:4873/@types/color-convert/-/color-convert-2.0.2.tgz#a5fa5da9b866732f8bf86b01964869011e2a2356"
```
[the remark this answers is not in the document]
AllanOricil: good you saw it. I use verdaccio and forgot to change haha

AllanOricil:
> > I can't really test the FLAG DESCRIPTIONS  section because I don't know any command that has flags without both summary and description. Do you know any?
> 
> `sf project deploy start` has a `FLAG DESCRIPTIONS` section

It works :D

image

src/config/config.ts
```diff
@@ -325,6 +336,14 @@ export class Config implements IConfig {
 
     this.npmRegistry = this.scopedEnvVar('NPM_REGISTRY') || this.pjson.oclif.npmRegistry
 
+    this.enableTheme = this.scopedEnvVarTrue('ENABLE_THEME') ?? this.pjson.oclif.enableTheme
```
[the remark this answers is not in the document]
AllanOricil: I liked it

src/config/config.ts
```diff
@@ -623,11 +629,6 @@ export class Config implements IConfig {
       .map((alias) => [alias.replaceAll('@', '').replaceAll(/[/-]/g, '_'), k].join('_').toUpperCase())
   }
 
-  public scopedEnvVarTrue(k: string): boolean {
```
mdonnalley, line 607: Removing this is a breaking change, which I'd like to avoid

What's the benefit of `scopedEnvVarBoolean`? Is it just so you can use the `??` syntax?

mdonnalley (changes requested):

AllanOricil:
@mdonnalley help me. Why isn't  my stub working?

image


I'm sure the path is the same, but the execution is still using the "original" method instead of resolving with my mocked data. 

image

src/config/config.ts
AllanOricil: Yes, it was to allow using it with ??
When a boolean env var isn't set we might also want to do something. I was using it, but after doing your recommended changes it is no longer necessary. I will revert these changes later.

mdonnalley:
@AllanOricil if you push up your latest changes I can take a look at the tests

AllanOricil:
> @AllanOricil if you push up your latest changes I can take a look at the tests

I think I found the issue. I was not assigning `test = test.stub...` i will finish these tests and the changes you asked. Then I push everything here and ask you revision again.

AllanOricil:
@mdonnalley did not work. Help me. My breakpoints are also no longer working when running `yarn test:debug`. When I attach a debugger, it simply is ignored. I don't really know what happened.

AllanOricil:
this line breaks my tests

image

AllanOricil:
I will call a wrap for today. Spent like 4 hours with no progress trying to mock a simple `this.safeReadJson`  method :(

mdonnalley:
@AllanOricil all the tests pass if you change the stub to

```
test.stub(fs, 'safeReadJson', (stub) => stub.resolves(theme))
```

AllanOricil:
> @AllanOricil all the tests pass if you change the stub to
> 
> ```
> test.stub(fs, 'safeReadJson', (stub) => stub.resolves(theme))
> ```

D: 

Really????? 😐
Wouldn't this cause all safeReadJson calls to return theme?

AllanOricil:
> @AllanOricil I think the failing `esm-cjs` test is because of one of your changes:
> 
> ```
>   e2e:esm-cjs.ts:cjs:plugin-test-cjs-1 stdout $ shx rm -rf dist && tsc -b
> node_modules/@oclif/core/lib/interfaces/config.d.ts(2,8): error TS1259: Module '"/private/var/folders/4l/6tqf4pm156j751spjj1qppwr0000gp/T/esm-cjs.ts/cjs/plugin-test-cjs-1/node_modules/@oclif/core/node_modules/@types/color/index"' can only be default-imported using the 'esModuleInterop' flag
> node_modules/@oclif/core/lib/help/util.d.ts(1,8): error TS1259: Module '"/private/var/folders/4l/6tqf4pm156j751spjj1qppwr0000gp/T/esm-cjs.ts/cjs/plugin-test-cjs-1/node_modules/@oclif/core/node_modules/@types/color/index"' can only be default-imported using the 'esModuleInterop' flag
> info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
> ```
> 
> That seems to be saying that the `colors` package cannot be default imported. Probably just needs to be changed to `import * as Color from 'color'`

I can't use `import * as Color from 'color'`

image

AllanOricil:
> > @AllanOricil I think the failing `esm-cjs` test is because of one of your changes:
> > ```
> >   e2e:esm-cjs.ts:cjs:plugin-test-cjs-1 stdout $ shx rm -rf dist && tsc -b
> > node_modules/@oclif/core/lib/interfaces/config.d.ts(2,8): error TS1259: Module '"/private/var/folders/4l/6tqf4pm156j751spjj1qppwr0000gp/T/esm-cjs.ts/cjs/plugin-test-cjs-1/node_modules/@oclif/core/node_modules/@types/color/index"' can only be default-imported using the 'esModuleInterop' flag
> > node_modules/@oclif/core/lib/help/util.d.ts(1,8): error TS1259: Module '"/private/var/folders/4l/6tqf4pm156j751spjj1qppwr0000gp/T/esm-cjs.ts/cjs/plugin-test-cjs-1/node_modules/@oclif/core/node_modules/@types/color/index"' can only be default-imported using the 'esModuleInterop' flag
> > info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That seems to be saying that the `colors` package cannot be default imported. Probably just needs to be changed to `import * as Color from 'color'`
> 
> I can't use `import * as Color from 'color'`
> 
> image

src/interfaces/config.ts
```diff
@@ -24,6 +26,54 @@ export type VersionDetails = {
...
+  'flagRequired',
+  'flagSeparator',
+  'flagType',
+  'sectionDescription',
+  'sectionHeader',
+  'topic',
+  'version',
+]
+
+export function parseTheme(untypedTheme: any): Theme {
```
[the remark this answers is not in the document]
AllanOricil: fixed

yarn.lock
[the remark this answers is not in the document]
AllanOricil: fixed

src/config/config.ts
AllanOricil: fixed

mdonnalley:
@AllanOricil 

Works for me if I do this in `interfaces/config.ts`

```
import * as Color from 'color'

export function parseTheme(untypedTheme: Record<string, string>): Theme {
  return Object.fromEntries(
    Object.entries(untypedTheme)
      .filter(([key]) => THEME_KEYS.includes(key))
      // eslint-disable-next-line new-cap
      .map(([key, value]) => [key, new Color.default(value)]),
  )
}
```

and this in `help/util.ts`

```
import * as Color from 'color'
```

src/config/config.ts
```diff
@@ -604,8 +620,8 @@ export class Config implements IConfig {
...
-    const v = process.env[this.scopedEnvVarKeys(k).find((k) => process.env[k]) as string]
+  public scopedEnvVarTrue(k: string): boolean | undefined {
```
mdonnalley, line 623: shouldn't this just be `boolean`?

src/interfaces/config.ts
```diff
@@ -24,6 +26,51 @@ export type VersionDetails = {
...
+  'flagRequired',
+  'flagSeparator',
+  'flagType',
+  'sectionDescription',
+  'sectionHeader',
+  'topic',
+  'version',
+]
+
+export function parseTheme(untypedTheme: Record<string, string>): Theme {
```
mdonnalley, line 66: This doesn't belong in the `interfaces` directory. Should go somewhere under `utils`

src/interfaces/config.ts
```diff
@@ -24,6 +26,51 @@ export type VersionDetails = {
...
+  return Object.fromEntries(
+    Object.entries(untypedTheme)
+      .filter(([key]) => THEME_KEYS.includes(key))
```
mdonnalley, line 69: I don't understand the value of this line. I understand that it's going to remove any un-supported theme key from the final result but why would it be bad to have unsupported keys in the final theme? They won't do any harm. It also would allow a future where the theme could be customized beyond just help. For instance, someone wants to have a theme for their CLI's log messages

Plus, if you remove it then you can remove the `THEME_KEYS` const that just duplicates the `Theme` type

But if you think it's absolutely imperative to filter the keys, I think this is a cleaner solution:

```typescript
export const THEME_KEYS = [
  'alias',
  'bin',
  'command',
  'commandSummary',
  'dollarSign',
  'flag',
  'flagDefaultValue',
  'flagOptions',
  'flagRequired',
  'flagSeparator',
  'flagType',
  'sectionDescription',
  'sectionHeader',
  'topic',
  'version',
]

export type Theme = Record<string, Color>

export function parseTheme(untypedTheme: Record<string, string>): Theme {
  return Object.fromEntries(
    Object.entries(untypedTheme)
      .filter(([key]) => THEME_KEYS.includes(key))
      // eslint-disable-next-line new-cap
      .map(([key, value]) => [key, new Color.default(value)]),
  )
}
```

mdonnalley (changes requested):

src/config/config.ts
AllanOricil: fixed

src/interfaces/config.ts
AllanOricil: fixed

src/interfaces/config.ts
AllanOricil: I will keep the filter to: 

- avoid wasting time parsing color codes that are never going to be used
- avoid storing colors in memory that are never going to be used 

I know it does not make any difference in performance since it is a really small number of props, but I want to leave it there anyway.

image

AllanOricil: fixed

AllanOricil:
> /@types/color/index"' can only be default-imported using the 'esModuleInterop' flag

fixed

src/interfaces/pjson.ts
```diff
@@ -25,6 +25,7 @@ export namespace PJSON {
       default?: string
       description?: string
       devPlugins?: string[]
+      enableTheme: boolean
```
mdonnalley, line 28: I don't think you need this anymore

src/interfaces/config.ts
```diff
@@ -117,11 +140,12 @@ export interface Config {
   scopedEnvVar(key: string): string | undefined
   scopedEnvVarKey(key: string): string
   scopedEnvVarKeys(key: string): string[]
-  scopedEnvVarTrue(key: string): boolean
+  scopedEnvVarTrue(key: string): boolean | undefined
```
mdonnalley, line 143: This also needs to be reverted

src/config/config.ts
```diff
@@ -325,6 +328,11 @@ export class Config implements IConfig {
 
     this.npmRegistry = this.scopedEnvVar('NPM_REGISTRY') || this.pjson.oclif.npmRegistry
 
+    const themeFilePath = resolve(this.configDir, 'theme.json')
+    const theme = this.scopedEnvVarTrue('DISABLE_THEME') ? undefined : await safeReadJson(themeFilePath)
```
mdonnalley, line 332:
```typescript
const theme = this.scopedEnvVarTrue('DISABLE_THEME') ? undefined : await safeReadJson<Record<string, string>>(themeFilePath)
```

That way you don't have to use `as` when passing it to `parseTheme`

mdonnalley (changes requested):
Just a few small things and then I'll approve.

src/config/config.ts
AllanOricil: fixed at 00a6ccdde08f13b44eec55b57a6d6e584f02d9fd

Could you tell me if you agree with what I did. I removed enableTheme from config.

AllanOricil:
Im really lazy to write tests to prove what each theme prop changes in the rendered helper message. Do you want to do it? I have created a test plan but I did not commit it.

mdonnalley:
@AllanOricil I'm going to merge this into a separate feature branch so that we can run all the integration tests before merging. I'm taking a vacation tomorrow so I would expect to get this merged into `main` early next week

AllanOricil:
@mdonnalley thank you for all the help 👍

PR #855, opened 2023-11-06
Add support to the NO_COLOR Manifest

AllanOricil:
Add support to the [NO_COLOR manifest](https://no-color.org/).

- When `NO_COLOR` is set, chalk won't display any colors `FORCE_COLORS=0`
- When `NO_COLOR` is unset, chalk displays all possible colors `FORCE_COLORS=3`

CLI developers no longer need to create their own env variable to remove ansi colors from their command outputs. You can then include Heroku and SFDX CLIs in the NO_COLOR manifest opening a PR in this repo https://github.com/jcs/no_color

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,9 @@
 import write from './cli-ux/write'
 
+// learn about the no-color manifest in this link: https://no-color.org/
+// learn about how to disable chalks colors in this link: https://github.com/chalk/chalk#supportscolor
+process.env.FORCE_COLOR = (Number(!process.env.NO_COLOR) * 3).toString()
+
 function checkCWD() {
   try {
     process.cwd()


mdonnalley:
@AllanOricil I like the intent of this but I don't think it's something we want to add at the moment. A few reasons:

- I agree with `chalk`'s rationale for not adding `NO_COLOR` support
  - https://github.com/chalk/supports-color/issues/105#issuecomment-542271445
  - https://github.com/chalk/supports-color/issues/74
- If we merge this we still couldn't say that we adhere to the NO_COLOR standard since we couldn't guarantee that all colors would be disabled - someone could have a CLI plugin that uses a different color library that doesn't respect either `FORCE_COLOR` or `NO_COLOR`
- In order to actually adhere to the standard oclif/core would need to know about every color related environment variable. Maybe that wouldn't be that difficult but it sets a bad precedence 
- I think this is a scenario where I prefer oclif to be un-opinionated and allow CLI developers to make up their own mind. Especially since it would be trivially easy for someone to set the env var themselves in their `bin/run.js`

AllanOricil:
I understand that there could exist plugins using other color libs, but wouldnt this mean that they are not conforming with Oclif's core? Shouldnt all plugins and CLIs use Oclif's UX object to log? If they are not using it, why care about these plugins/CLIS? They chose not to use oclif core packages for outputs. It is their problem, and oclifs for allowing them to not follow the "framework".

I also can't understand why not enforcing this on ALL CLIs and plugins made with oclif's core.  Leaving colors optional for cli builders just contributes to people not following the NO_COLOR manifest.

If all tools out there that are used to build CLIs enforce NO_COLOR, over the years there would exist less need for a NO_COLOR manifest.

AllanOricil:
obs: I did not check if this information is true. There may exist other libraries that were not included.

According to chat gpt these are the most used libs that developers use to style their terminal outputs with ansi colors: 

| **Package Name** | **Link** |
|------------------|----------|
| Kleur            | `https://www.npmjs.com/package/kleur` |
| Colorette        | `https://www.npmjs.com/package/colorette` |
| Colors           | `https://www.npmjs.com/package/colors` |
| Blessed          | `https://www.npmjs.com/package/blessed` |

AllanOricil:
Ok, I give up on this. According to chalk's maintainer, there are already lots of different "standards"

PR #859, opened 2023-11-09
feat: add flag type label

AllanOricil:
This will enhance the way (flags | args) are displayed. Developers will no longer need to read the flag's description to grasp the type|format of value they have to use with a certain flag or arg. For example, when I read `filePath` I immediately understand that a certain flag require a file path.

Below you can see how docker cli does it

image

obs: I added a new property called `typeLabel` instead of using the `helpValue` because I did not want to force cli developers to use it, and because the variable name better describe what it is than helpValue

## BEFORE

image

## AFTER

To enable this render type CLI developers have to set `oclif.showFlagTypeLabel` to `true` in the `package.json`

image

When `typeLabel` is not defined, and `oclif.showFlagTypeLabel` is true instead of showing `=<option>` it is displayed `string`, or `string...` if multiple. It feels cleaner in my opinion.

image

----

## When combining with the theme and indent features

image

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -91,6 +91,7 @@ export class Config implements IConfig {
   public plugins: Map<string, IPlugin> = new Map()
   public root!: string
   public shell!: string
+  public showFlagTypeLabel!: boolean
   public topicSeparator: ' ' | ':' = ':'
   public userAgent!: string
   public userPJSON?: PJSON.User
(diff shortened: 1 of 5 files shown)


mdonnalley:
@AllanOricil Thanks for making this PR but I don't think it's something we want to move forward with. I have two main concerns:

1. I think it makes the help output too crowded
2. If I understand correctly, `typeLabel` would need to be added to every single flag. Otherwise, it would show something like `--integer string` (as an example), which is confusing.  So I'm not sure how it's better than the existing `helpValue` property

AllanOricil:
@mdonnalley my idea was to deprecate the other prop and eventually merge both into one. This way developers can decide what to show in their flags. In my opinion, I really think that showing `directoryPath` instead of `<value>` offers a much better dx. I also think the variable name is better than the current one.

PR #861, opened 2023-11-09
fix: add exit codes to different flag validation errors

WillieRuemmele:
@W-14408666@

adds different exit codes (3-7) for different flag/args validation errors


mdonnalley (changes requested):
@WillieRuemmele I think this needs to be configurable at the CLI level so we don't force all CLIs to use our standards. I'd like to keep oclif as un-opinionated as possible

I'm thinking that the CLI could have something like this in the package.json

```json
{
  "oclif": {
    "exitCodes": {
        "invalidArgs": 3,
        "requiredFlag": 4,
        // etc...
      }
  }
}
```

You might have an issue with accessing the root plugin's package.json from `parser/validate.ts`. If so, you could consider putting all the exit codes on the `Cache` class and then access them there - that way you don't have to pass around the package.json through multiple levels

mshanemc (dismissed):
approved once comment stuff is cleaned up.

mshanemc:
QA notes:  

build oclif/core and `yarn link` into the local CLI (from installer) `/Users/shane.mclaughlin/.local/share/sf/client/current`

run a command, then `echo $?` to see exitCode.
`sf version --foo` => 2
saying "no" to did-you-mean after entering a non-existent command => 127
`sf org create sandbox` (missing flag) => 1
`sf org list --no-prompt` (missing dependsOn) => 1
 
I *think* these are thrown from "inside" a command instead of during parse/validate
`sf update --version 0.0.0` => 1
`sf org display -o bad` => 1

---
and then we go customize the exitCodes prop 


`"nonExistentFlag": 404,` ❓   sf version --foo => 148 [I don't know how it gets to 148 from 404.  I think it's modulo 255?
`"nonExistentFlag": 8,` ✅  sf version --foo => 8 
`"nonExistentFlag": 33,` ✅  sf version --foo => 33

`"failedFlagValidation": 5,` `sf org list --no-prompt` (missing dependsOn) => 1 ❌ should have been 5.
[I tried a bunch of things and could never get `failedFlagValidation` to fire from `dependsOn`]

mdonnalley:
> `"nonExistentFlag": 404,` ❓ sf version --foo => 148 [I don't know how it gets to 148 from 404. I think it's modulo 255?

Turns out that exit codes beyond 255 aren't supported: https://www.shellscript.sh/exitcodes.html

mdonnalley:
> "failedFlagValidation": 5, sf org list --no-prompt (missing dependsOn) => 1 ❌ should have been 5.
[I tried a bunch of things and could never get failedFlagValidation to fire from dependsOn]

This is an issue with any command extending SfCommand. I'll have to fix that in a separate PR

mshanemc (approved):

PR #862, opened 2023-11-09
feat: enable themes (#852)

mdonnalley:
Merge #852 to `main`


AllanOricil:
I forgot to change http to https on those dependency entries

AllanOricil:
@mdonnalley when you release this feature could you please make some sort of "promotion" with my name? Now that I have time, I'm working on open source projects that I have interest in order to increase my chances of leaving this s*** country when my parents are gone. If you can't do it, there is no problem. Thank you in advance.

mdonnalley:
> when you release this feature could you please make some sort of "promotion" with my name?

@AllanOricil No problem. You're name will be in the commit history and we'll mention your name in the `sf` release notes once we integrate this feature there. Does that work for you?

AllanOricil:
@mdonnalley I would like to also have my name somewhere on those posts/videos that Salesforce usually release to notify devs about new features

AllanOricil:
@mdonnalley you removed the test that tests that a key is present in the parsed theme. If a developer remove `alias` from Theme, will all test pass right? Those tests were there to prove that the key can not vê removed from Theme to not cause a regression

mdonnalley:
@AllanOricil Are you referring to [these tests](https://github.com/oclif/core/pull/852/files#diff-e45f6f7133ba8e91ef7ca9031c5ee597721c772204d567b6f3a2fdfe1e4d1b11R170)?

If so, I don't see what value that provides that [this test](https://github.com/oclif/core/pull/852/files#diff-e45f6f7133ba8e91ef7ca9031c5ee597721c772204d567b6f3a2fdfe1e4d1b11R144) doesn't. They're the exact same tests except one has more properties in the parsed theme

AllanOricil:
> @AllanOricil Are you referring to [these tests](https://github.com/oclif/core/pull/852/files#diff-e45f6f7133ba8e91ef7ca9031c5ee597721c772204d567b6f3a2fdfe1e4d1b11R170)?
> 
> If so, I don't see what value that provides that [this test](https://github.com/oclif/core/pull/852/files#diff-e45f6f7133ba8e91ef7ca9031c5ee597721c772204d567b6f3a2fdfe1e4d1b11R144) doesn't. They're the exact same tests except one has more properties in the parsed theme

@mdonnalley in the first, if you remove `alias` from Theme, the code will not compile with a type error. In the second, because there is no ref to the key name, the code compiles. The benefit is that a dev will see that removing `alias` from Theme causes a regression and he can must not do it

AllanOricil:
@mdonnalley look, when I removed `alias` from THEME_KEYS, 1 single test failed so that a developer can know exactly the root cause of the problem. It is like if it was a real world experiment, you change 1 variable, it changes 1 single behaviour.

image

AllanOricil:
@mdonnalley  Yes, with the current code there is no way to know if another developer deleted any of the mandatory Theme keys because there is no tests to test the which keys are valid in the Theme specification.

src/interfaces/theme.ts
```diff
@@ -0,0 +1,43 @@
...
+  'bgRedBright',
+  'bgGreenBright',
+  'bgYellowBright',
+  'bgBlueBright',
+  'bgMagentaBright',
+  'bgCyanBright',
+  'bgWhiteBright',
+] as const
+
+export type StandardChalk = (typeof STANDARD_CHALK)[number]
```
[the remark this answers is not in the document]
AllanOricil: Never mind, I tested and there are some colors, like 'blackBright' that can't be parsed by the color package. Really good finding!

test/util/theme.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      flag: 'cyan',
+      flagDefaultValue: 'cyan',
+      flagOptions: 'cyan',
+      flagRequired: 'cyan',
+      flagSeparator: 'cyan',
+      sectionDescription: 'cyan',
+      sectionHeader: 'cyan',
+      topic: 'cyan',
+      version: 'cyan',
+    }
```
AllanOricil, line 81: `cyan` is supported by the colors package. Use a color, like `blackBright`, that would fail to be parsed without `STANDARD_CHALK` in order to make this a non false positive test.

Here is a test where I discovered that `blackBright` isn't supported by the color package
image

src/util/theme.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+}
+
+function isStandardChalk(color: any): color is StandardChalk {
+  return STANDARD_CHALK.includes(color)
+}
+
+export function colorize(color: Color | StandardChalk | undefined, text: string): string {
+  if (isStandardChalk(color)) return standardChalk(color, text)
```
AllanOricil, line 156: I think you can remove this function and just use `return chalk[color](text)`

Using a theme with standard chalk colors

image

src/util/theme.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  return color ? chalk.hex(color.hex())(text) : text
+}
+
+export function parseTheme(untypedTheme: Record<string, string>): Theme {
+  return Object.fromEntries(Object.entries(untypedTheme).map(([key, value]) => [key, getColor(value)]))
```
AllanOricil, line 162: the theme feature should not waste time or memory storing tokens its <b>spec</b> does not support (strict). Its spec supports these tokens:

````js
export const THEME_KEYS = [
  'alias',
  'bin',
  'command',
  'commandSummary',
  'dollarSign',
  'flag',
  'flagDefaultValue',
  'flagOptions',
  'flagRequired',
  'flagSeparator',
  'flagType',
  'sectionDescription',
  'sectionHeader',
  'topic',
  'version',
]
````

Then you have to add these [tests](https://github.com/AllanOricil/core/blob/00a6ccdde08f13b44eec55b57a6d6e584f02d9fd/test/util/util.test.ts#L170-L327) back because they prove the independency and usage of each token of Theme Spec. We need it so that when tokens are changed we know exactly the regressions the change made. For example, if I remove the `alias` token and run the tests, I will see that its test failed, and this will show me a regression:

image

**Review by AllanOricil — CHANGES_REQUESTED** [review]

File: src/util/theme.ts (hunk 4 above)
Comment (AllanOricil) [reply src/util/theme.ts:162]: @mdonnalley we would change `logToStderr` to the following piece of code so that they can keep using `this.logToStderr` without updating their code. The theme.json is what controls the color changing.

````js
  public static logToStderr(format?: string, ...args: string[]): void {
    write.stderr(colorize(this.config?.theme?.warningMsg, utilFormat(format, ...args) + '\n'))
  }
````

and a new token would be added to the Theme Spec

````js
const THEME_KEYS = {
  ...,
  "warningMsg"
}

````

Comment (AllanOricil) [reply src/util/theme.ts:162]: If `this.config?.theme?warningMsg` does not exist, `this.logToStderr` will output what is currently outputing. It would not require any code updates in CLIs using @oclif/core.

Comment (AllanOricil) [reply src/util/theme.ts:162]: If CLI developers want to overwrite what the theme.json says, they can do this

````js
  public async run(): Promise<void> {
    this.config.theme.warningMsg = getColor('brightBlack')
    this.logToStderr('hello world')
  }
}
````

Comment (AllanOricil) [reply src/util/theme.ts:162]: and for `logToStderr`, a better Theme Key is `stderrMessage`,

````js
  public static logToStderr(format?: string, ...args: string[]): void {
    write.stderr(colorize(this.config?.theme?.stderrMessage, utilFormat(format, ...args) + '\n'))
  }
````

Comment (mdonnalley) [reply src/util/theme.ts]: It's a very poor assumption that everyone would be using `logToStderr` to only log warning messages. But my point was that by allowing any key in `themes.json` we give CLI developers the ability to create a more extensive theme for their users. I have no interest in standardizing themes for every single oclif CLI

File: src/interfaces/theme.ts [hunk 5]
```diff
@@ -40,4 +40,41 @@ export const STANDARD_CHALK = [
...
+  'sectionDescription',
+  'sectionHeader',
+  'topic',
+  'version',
+] as const
+
+export type ThemeKey = (typeof THEME_KEYS)[number]
+
+export type Theme = {
+  [key: string | ThemeKey]: Color | StandardChalk | undefined
```
AllanOricil, line 64: Ok, I see a use case for this. CLI developers can extend the oclif's Theme Spec with their own theme specific keys so that plugins built with their own CLI API can take advantage of their own Theme Spec, which is just a combination of Oclif + CLI Theme Specs. Makes sense now.

AllanOricil (approved):

src/interfaces/theme.ts
AllanOricil: Shouldn't Theme be a `interface` to allow it to be extensible in typescript world?

src/util/theme.ts
AllanOricil: Going back to what I said earlier, there are no tests to "prove" the Theme Spec the Core Specifies... 

If the core is using certain `keys` everywhere in its code, there must exist tests to prove each single one of these keys....

AllanOricil: ok, thanks for adding a test back. Now it is working again as I specified.

mdonnalley: That's what this is: https://github.com/oclif/core/pull/862/files#diff-6bef7aac23bf8d9e00a172402d3b443a7a6e1834198135027ce593fa877e8f73R116

src/interfaces/theme.ts
```diff
@@ -0,0 +1,85 @@
...
+  'bgBlueBright',
+  'bgMagentaBright',
+  'bgCyanBright',
+  'bgWhiteBright',
+  'bold',
+  'underline',
+  'dim',
+  'italic',
+  'strikethrough',
+] as const
```
AllanOricil, line 44: Can't these exports work? I did not check their values, but by looking at the variable names it seems to be what we need

https://github.com/chalk/chalk/blob/main/source/index.js#L207-L218

AllanOricil: Never mind. That is for v5. V4 does not have those exports, unfortunately

AllanOricil: In v4 you could use  



````js
Object.getOwnPropertyNames(chalk.prototype)
````

Based on the following, all styles are direct props from the chalk prototype:

https://github.com/chalk/chalk/blob/95d74cbe8d3df3674dec1445a4608d3288d8b73c/source/index.js#L222

File: src/config/config.ts [hunk 7]
```diff
@@ -390,6 +399,23 @@ export class Config implements IConfig {
...
+      : resolve(this.configDir, 'themes.json')
+    const themes = await safeReadJson<Themes>(themesFile)
+    const activeTheme = themes ? parseTheme(themes) : undefined
+    return {
+      activeTheme,
+      file: themesFile,
+      themes,
+    }
+  }
+
```
AllanOricil, line 418: I don't understand why loading all themes in memory if only one is used at a time

AllanOricil: It does not answer the question. What is the use case to have all themes in memory?

AllanOricil:
I envisioned and started this feature and you took it over and started to make decisions without discussing with me. This is extremely unprofessional and it will be registered here for other people to see.

mdonnalley:
> I envisioned and started this feature and you took it over and started to make decisions without discussing with me. This is extremely unprofessional and it will be registered here for other people to see.

It's an open PR - no decisions have been made. Your feedback is welcome, including on the plugin. 

Your reviews and comments on this PR have been incredibly helpful thus far - I'm sure you've noticed that many of the commits here are addressing your feedback. From my perspective, this has been a collaborative effort.

The current implementation in this PR and the plugin-theme PR was discussed [here](https://github.com/oclif/core/discussions/871#user-content-fn-1-f88e9fc596de913edf2d1d2bbfee10e2). As Cristian mentioned in his comment, we're not willing to take on the complexity of having themes be installable via npm so we're looking for a more file-based solution. We'd love if you could add your thoughts there so that we can starting working towards a solution that we're all happy with

AllanOricil:
Hi @mdonnalley sorry for what I said earlier. From time to time I have stupid emotional outages and I end up saying stupid things because my mind starts to think about things that doesnt exist. For example, I think that Im losing something and being left behind, or I get envy, again for no reason. I take some medication but it just doesn't work well sometimes. You can continue this and follow your ideas. You actually did most of the good decisions. I hope to use it when it is finished. Again, sorry.

Issue #496, opened 2022-09-22
argument not read, if after a flag with multiple: true


shazron:
**Describe the bug**
I have a command that has 1 arg, and 1 multiple flag. If I specify the arg after the flag, I cannot access the arg in my command.

**To Reproduce**
```javascript
const { Command, Flags } = require('@oclif/core')

class FlagsTestCommand extends Command {
  async run () {
    const { args, flags } = await this.parse(FlagsTestCommand)

    this.log('args.path', args.path)
    this.log('flags.myflags', flags.myflags)
  }
}
FlagsTestCommand.description = `Create a new Adobe I/O App
`

FlagsTestCommand.flags = {
  myflags: Flags.string({
    description: 'my flag for testing',
    char: 'm',
    multiple: true
  })
}

FlagsTestCommand.args = [
  {
    name: 'path',
    description: 'Path to the test directory',
    default: '.'
  }
]

module.exports = FlagsTestCommand
```

run:
1. `oclif FlagsTest -m=abc123 Foobar` (no args parsed)
2. `oclif FlagsTest -m abc123 Foobar` (no args parsed)
3. `oclif FlagsTest Foobar -m=abc123 ` (arg is parsed since it occurs before the flag)

**Expected behavior**
I can access my arg if specified after the multiple flag.

**Actual behavior**
I cannot access my arg if specified after the multiple flag.

**Workaround**
Specify the arg before the multiple flag. 

**Other context**
AFAIK you can put args before flags, or after. 
`An argument is anything to the right of a command that is not a flag. An argument can come before or after flags.` from https://oclif.io/blog/2019/02/20/cli-flags-explained

mdonnalley:
@shazron Thanks for posting the issue. I validated that this is in fact a bug. However, the fix will require that we make a breaking change to the flag parser.

Right now both of the following are valid ways of providing multiple flag values:

* `my-command --multi-flag 1 --multi-flag 2 --multi-flag 3`
* `my-command --multi-flag 1 2 3`

Supporting the second way is what's causing this issue. The flag parser will assume that each subsequent value is part of the flag until the next flag is found. So in your case, it thinks that the argument is part of the value for the multiple flag.

To fix it, we'd have to drop support for that second method, which would be a breaking change. 

I've labeled this as `will-fix-in-v4` so that we can pick it up once we get started on the next major version. Until then, you'll have to provide the argument first before providing the values for the multiple flag.

mdonnalley:
[...]

PR #880, opened 2023-11-21
feat: add 'multipleNonGreedy' flag option to assign only one value per multiple flag

royra:
This is a port of https://github.com/oclif/parser/pull/79 by @cxam

Original description:

This feature aims to address the issue outlined in https://github.com/oclif/oclif/issues/190 to only allow a single value to be assigned for each flag that has the `multiple` option set. It's a non-breaking change that requires you to include the `singleValue` boolean in your flag options.

In the command `cmd -a foo bar`, it's generally accepted that `bar` becomes a positional argument and should be the default behaviour. Unfortunately, the [PR](https://github.com/oclif/parser/pull/25) that broke default behaviour has been part of this project for a while now and I feel providing the user an option to change this back would be the best course of action unless a breaking release is planned for oclif.

If this goes through, I'll update and push the relevant changes to the [docs repository](https://github.com/oclif/oclif.github.io).

### DIFF
--- a/src/interfaces/parser.ts
+++ b/src/interfaces/parser.ts
@@ -209,6 +209,12 @@ export type OptionFlagProps = FlagProps & {
   options?: readonly string[]
   multiple?: boolean
   /**
+   * Parse one value per flag; allow `-m val1 -m val2`, disallow `-m val1 val2`.
+   * Set to true to use "multiple: true" flags together with args.
+   * Only respected if multiple is set to true.
+   */
+  multipleNonGreedy?: boolean
+  /**
    * Delimiter to separate the values for a multiple value flag.
    * Only respected if multiple is set to true. Default behavior is to
    * separate on spaces.
--- a/src/parser/parse.ts
+++ b/src/parser/parse.ts
@@ -200,7 +200,7 @@ export class Parser<
         }
       }
 
-      if (parsingFlags && this.currentFlag && this.currentFlag.multiple) {
+      if (parsingFlags && this.currentFlag && this.currentFlag.multiple && !this.currentFlag.multipleNonGreedy) {
         this.raw.push({flag: this.currentFlag.name, input, type: 'flag'})
         continue
       }
--- a/test/parser/parse.test.ts
+++ b/test/parser/parse.test.ts
@@ -631,6 +631,48 @@ See more help with --help`)
       })
     })
 
+    describe('multiple flags with single value', () => {
+      it('parses multiple flags with single value', async () => {
+        const out = await parse(['--bar', 'a', 'b', '--bar=c', '--baz=d', 'e'], {
+          args: {argOne: Args.string()},
+          flags: {
+            bar: Flags.string({multiple: true, multipleNonGreedy: true}),
+            baz: Flags.string({multiple: true}),
+          },
+        })
+        expect(out.flags.baz?.join('|')).to.equal('d|e')
+        expect(out.flags.bar?.join('|')).to.equal('a|c')
+        expect(out.args).to.deep.equal({argOne: 'b'})
+      })
+
+      it('parses multiple flags with single value multiple args', async () => {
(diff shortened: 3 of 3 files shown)


mdonnalley:
@royra Thanks for the PR! Can you help me understand why someone would use `multiple: true` and `singleValue: true`? Seems to me that they would just negate each other - why not just have a normal flag that doesn't set `multiple: true`?

royra:
> Can you help me understand why someone would use `multiple: true` and `singleValue: true`?

This solves the problem with combining `{ multiple: true }` flags with args, as described in https://github.com/oclif/oclif/issues/190:

Example: assuming `m` is a `{ multiple: true }` flag and an arg `argOne` is defined:
```bash
command -m val arg1
```
should parse into 
```javascript
{ 
  flags: { m: ['val'] }, 
  args: { argOne: 'arg1' },
}
```

but instead parses `arg1` as the 2nd value of `m`.

I simply ported the original PR, and I agree `singleValue` isn't very descriptive. Maybe `multipleNonGreedy`?

royra:
> Can you help me understand why someone would use `multiple: true` and `singleValue: true`?

This solves the problem with combining `{ multiple: true }` flags with args, as described in https://github.com/oclif/oclif/issues/190:

Example: assuming `m` is a `{ multiple: true }` flag and an arg `argOne` is defined:
```bash
command -m val arg1
```
should parse into 
```javascript
{ 
  flags: { m: ['val'] }, 
  args: { argOne: 'arg1' },
}
```

but instead parses `arg1` as the 2nd value of `m`.

I simply ported the original PR, and I agree `singleValue` isn't very descriptive. Maybe `multipleNonGreedy`?

royra:
Related: https://github.com/oclif/core/issues/496

mdonnalley:
@royra Thanks but I'm still not sure I understand how this new property solves anything

I understand the problem though. You want to be able to write a command that takes arguments and multi flags and have it parsed properly.

But if you're adding a new property to force the parser to stop looking for flag values after it finds the first one, then why not just define the flag as non multi flag? As I understand it these two flag definitions are the exactly the same:

```typescript
flag1: Flags.string({
  multiple: true,
  singleValue: true,
})
```

```typescript
flag1: Flags.string()
```

Because of that, I'm not inclined to move forward with this PR as it is. 

My recommendations are:
- Avoid mixing arguments and multi flags (this was my [recommendation]((https://github.com/oclif/core/issues/496#issuecomment-1377913588) in #496)
- If you must support mutli flags with arguments, you could create a custom flag that accepts an array of comma separated values:
```
export const arrayFlag = Flags.custom<string[]>({
  delimiter: ',',
})
```

royra:
> @royra Thanks but I'm still not sure I understand how this new property solves anything
> 
> I understand the problem though. You want to be able to write a command that takes arguments and multi flags and have it parsed properly.
> 
> But if you're adding a new property to force the parser to stop looking for flag values after it finds the first one, then why not just define the flag as non multi flag? As I understand it these two flag definitions are the exactly the same:
> 
> ```ts
> flag1: Flags.string({
>   multiple: true,
>   singleValue: true,
> })
> ```
> 
> ```ts
> flag1: Flags.string()
> ```
> 

This definition doesn't allow for specifying the flag more than once, e.g, `cmd --m val1 --m val2` - it results the error `Flag --m can only be specified once`.

Specifying a flag multiple times is a common pattern; unix classics like `sed` and `find` allow it, and modern CLIs like docker too.  Having an option for comma-separated values is a useful alternative, and I would like to have the option to parse both. I think this fix is a good compromise, since it allows this while not breaking backwards compat.

royra:
> In that case, I'd like for `singleValue` to be renamed to something more clear. You're earlier suggestion of `multipleNonGreedy` works for me. I also like `multipleStrict`, `multipleExplicit` - whatever you like best

@mdonnalley Done. Thank you!

mdonnalley (approved):

PR #886, opened 2023-11-26
fix: command not found on empty arg

royra:
This PR fixes an edge case bug in command ID parsing when the topic separator is a space

When `argv` contains an empty string arg, the command ID extracted contains an extra colon.

An empty string arg can occur when running the CLI as part of a script.

Example:

```bash
set FOO=""
cli cmd ${FOO} --bar
```

Result: command not found "cmd"

Root cause: incorrect parsing in `standardizeIDFromArgv`, which for the above example returned `cmd:` as the command ID.

### DIFF
--- a/src/cli-ux/README.md
+++ b/src/cli-ux/README.md
@@ -180,7 +180,7 @@ export default class Users extends Command {
         extended: true
       }
     }, {
-      printLine: this.log,
+      printLine: this.log.bind(this),
       ...flags, // parsed flags
     })
   }


mdonnalley:
@royra This looks like the same changes as https://github.com/oclif/core/pull/880 - is that intentional?

royra:
> @royra This looks like the same changes as #880 - is that intentional?

@mdonnalley you're right! I opened this PR with the wrong branch by mistake. when I tried to fix it just now it closed, so I reopened as #887. sorry for the mixup.

PR #887, opened 2023-11-29
 fix: command not found on empty arg

royra:
(this is a duplicate of #886 which had the wrong branch and was closed)

This PR fixes an edge case bug in command ID parsing when the topic separator is a space

When `argv` contains an empty string arg, the command ID extracted contains an extra colon.

An empty string arg can occur when running the CLI as part of a script.

Example:

```bash
set FOO=""
cli cmd ${FOO} --bar
```

Result: command not found "cmd"

Root cause: incorrect parsing in `standardizeIDFromArgv`, which for the above example returned `cmd:` as the command ID.

### DIFF
--- a/src/help/util.ts
+++ b/src/help/util.ts
@@ -23,7 +23,7 @@ function collateSpacedCmdIDFromArgs(argv: string[], config: IConfig): string[] {
 
     const final: string[] = []
     const idPresent = (id: string) => ids.has(id)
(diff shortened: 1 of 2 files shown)


mdonnalley (approved):

mdonnalley:
@royra I merged this into a feature branch (#887) so that all the integration tests (which don't run on forks) can run first. Once they pass I'll merge that branch and this will be available. Thanks for the contribution!

PR #890, opened 2023-11-30
feat: support tsx

mdonnalley:
Developers can change their `bin/dev.js` to use `tsx` instead of `node` or `ts-node` and everything will work as expected without this change. This PR simply avoids unnecessary operations related to `ts-node` that aren't necessary when executing with `tsx`


mdonnalley:
Closed in favor of #966

Issue #761, opened 2023-08-15
Allow flags to read its value from stdin


cristiand391:
**Is your feature request related to a problem? Please describe.**
Logging this request after discussing this PR with the team:
https://github.com/salesforcecli/plugin-auth/pull/765

**Describe the solution you'd like**
Allow flags to get their value from stdin if their value is `-`.

This is a common pattern in CLIs, `oclif` supports args reading from stdin but not flags.
> If input or output is a file, support - to read from stdin or write to stdout. This lets the output of another command be the input of your command and vice versa, without using a temporary file.

https://clig.dev/#arguments-and-flags

Design:
off by default, flag needs `allowStdin: true` and value `-` to read from stdin:
```ts
'sfdx-url-file': Flags.file({
      char: 'f',
      allowStdin: true,
      summary: messages.getMessage('flags.sfdx-url-file.summary'),
      required: true,
      deprecateAliases: true,
      aliases: ['sfdxurlfile'],
    })
```

```bash
# reads from stdin
echo $url | sf org login sfdx-url --sfdx-url-file -

# reads from file
sf org login sfdx-url --sfdx-url-file auth.json
``` 

**Describe alternatives you've considered**
We considered replacing `--sfdx-url-file` with an arg but decided this feature would:
1) allow us to follow our style guide (prefer flags over args)
2) benefit other CLIs

cristiand391:
From Shane:
> You'd still have to have oclif prevent people from doing - on more than 1 flag.
This can be mentioned in `allowStdin` description but preventing this from happening in the parser is out of the scope of this request and it's up to authors to ensure only 1 flag has this enabled.

💡 : this could be prevented via eslint rules, maybe we should start doing oclif-specific ones.

cristiand391:
`kubectl apply` example:
https://kubernetes.io/docs/reference/kubectl/cheatsheet/#creating-objects
```
# Create multiple YAML objects from stdin
kubectl apply -f - <<EOF
apiVersion: v1
kind: Pod
...
```

AllanOricil:
Some flags could be stdin only.

AllanOricil:
Is anybody working on this? If not I'm going to try

AllanOricil:
What is the plan to pass 2 flags values through stdin?

AllanOricil:
I think there should exist only one `--*-stdin` flag at a time. So if there are 2 flags with `allowStdin`, Users can't use both at the same time.

mdonnalley:
[...]

PR #894, opened 2023-12-05
feat: allow stdin option for flags

k-capehart:
- Add option for flags: allowStdin
- If a value of '-' is provided, then read from stdin
- Based on solution provided in https://github.com/oclif/core/issues/761
- Does not solve for the situation in which a developer creates multiple flags with the allowStdin property
    - Based on above chatter, this should be up to the developer to provide a good experience

### DIFF
--- a/src/parser/parse.ts
+++ b/src/parser/parse.ts
@@ -122,6 +122,8 @@ export class Parser<
   public async parse(): Promise<ParserOutput<TFlags, BFlags, TArgs>> {
     this._debugInput()
 
+    let stdinRead = false
+
     const parseFlag = (arg: string): boolean => {
       const {isLong, name} = this.findFlag(arg)
       if (!name) {
@@ -155,6 +157,11 @@ export class Parser<
           throw new CLIError(`Flag --${name} expects a value`)
         }
 
+        if (flag.allowStdin && input === '-') {
+          stdinRead = true
+          return true
+        }
+
         this.raw.push({flag: flag.name, input, type: 'flag'})
       } else {
         this.raw.push({flag: flag.name, input: arg, type: 'flag'})
@@ -182,6 +189,14 @@ export class Parser<
         }
 
         if (parseFlag(input)) {
+          if (parsingFlags && this.currentFlag && stdinRead) {
+            let stdin = await readStdin()
+            if (stdin) {
+              stdin = stdin.trim()
+              this.raw.push({flag: this.currentFlag.name, input: stdin, type: 'flag'})
+            }
+          }
+
           continue
         }
 
--- a/test/parser/parse.test.ts
+++ b/test/parser/parse.test.ts
@@ -1,11 +1,13 @@
 import {assert, config, expect} from 'chai'
+// eslint-disable-next-line node/no-extraneous-import
+import mockStdin from 'mock-stdin'
 import * as fs from 'node:fs'
 import {URL} from 'node:url'
 import {SinonStub, createSandbox} from 'sinon'
 
 import {Args, Flags} from '../../src'
 import {CLIError} from '../../src/errors'
-import {FlagDefault} from '../../src/interfaces/parser'
+import {FlagDefault, OutputArgs, OutputFlags, ParserOutput} from '../../src/interfaces/parser'
 import {parse} from '../../src/parser'
 
 config.truncateThreshold = 0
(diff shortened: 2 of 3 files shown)


test/parser/parse.test.ts
```diff
@@ -1855,3 +1857,40 @@ See more help with --help`)
     })
   })
 })
+
+describe('stdin', () => {
```
AllanOricil, line 1861: Maybe you should add a test to verify that myFlag has the value of stdin? 


````js
...
const { flags } = this.parse(cmd)

assert(flags.myFlag).equals(stdinValue)
...
````

Comment (k-capehart) [reply test/parser/parse.test.ts:1861]: I added an extra line in this test. Does that help?

`expect(out.flags.myflag).to.equals(stdinValue)`

File: test/parser/parse.test.ts [hunk 2]
```diff
@@ -1855,3 +1857,42 @@ See more help with --help`)
...
+
+  describe('allow stdin', () => {
+    beforeEach(() => {
+      execute()
+      stdin.send(stdinValue)
+      stdin.end()
+    })
+
+    it('should read stdin as input for command', () => {
+      afterEach(() => {
```
AllanOricil, line 1892: Why did u wrap these tests with afterEach?

AllanOricil:
@k-capehart what if you add a warning somewhere during compilation step to warn developers not use allowStdin more than once?

test/parser/parse.test.ts
k-capehart: I ran into some issues with the asserting executing before the stdin was read in, due to the parsing method being asynchronous. This is how I was able to force it to wait.

mdonnalley:
> > @k-capehart what if you add a warning somewhere during "compilation" (I forgot the proper word) step to warn developers not use allowStdin more than once?
> 
> I'm open to this, and can start looking into it. It could also be a follow up PR while we just update documentation for now.

@AllanOricil @k-capehart I don't think this provides enough value to warrant the complexity. We can always add it later if it's actually needed

src/parser/parse.ts
```diff
@@ -182,6 +189,14 @@ export class Parser<
         }
 
         if (parseFlag(input)) {
+          if (parsingFlags && this.currentFlag && stdinRead) {
```
mdonnalley, line 192: `parseFlag` is responsible for pushing to `this.raw`, so I think it'd be cleaner to move this processing to the `if` block you added on line 160

test/parser/parse.test.ts
mdonnalley: Does making the `it` callback `async` help this? This is what I have in mind:

```typescript
describe('allowStdin', () => {
  it('should read stdin as input for command', async () => {
    const stdin = mockStdin.stdin()
    stdin.send(stdinValue)
    stdin.end()
    const out = await execute()
    expect(out.raw[0].input).to.equal(stdinValue)
    expect(out.flags.myflag).to.equals(stdinValue)
  }
})
```

I didn't test that by the way. Just how I would attempt to do it myself.

You could also consider mocking the `readStdin` function in `src/parser/parse.ts` instead of trying to mock `stdin` directly

mdonnalley (changes requested):

src/parser/parse.ts
mdonnalley: You could make `parseFlag` async. Looks like it's only used once so it should be easy to do

test/parser/parse.test.ts
k-capehart: I just tried this. I still can't get it to work without using `afterEach()`. Depending on what I try it will either:

1. Throw an error that `out.raw is undefined` because it is still waiting for `readStdin()` to finish
2. Fail the assertion `expected '-' to equal 'x'` because something about making the `it` callback async is causing it to **not** want to wait for `readStdin()` to finish

k-capehart: I'm happy to try anything else though, let me know if the current approach is a major concern or bad practice.

mdonnalley: My main concern is that it's hiding some sort of race condition that'll come back to bite us later. If I were writing this I would stub `readStdin` instead of trying to use `mock-stdin`. Something like this:

```typescript
import {SinonSandbox, createSandbox} from 'sinon'
import * as parser from '../../src/parser'

describe('allowStdin', () => {
  let sandbox: SinonSandbox

  beforeEach(() => {
    sandbox = createSandbox()
  })

  afterEach(() => {
    sandbox.restore()
  })

  it('should read stdin as input for flag', async () => {
    sandbox.stub(parser, 'readStdin', 'x')
    const result = await parse(['--myflag', '-'], {
      flags: {
        myflag: Flags.string({allowStdin: true}),
      },
    })

   // assertions here
  })
})
```

k-capehart: @mdonnalley How does that look now?

mdonnalley (approved):

mdonnalley:
@k-capehart Thanks for doing this work 🏆 

I'm going to merge this into a separate feature branch so that all our integration tests (which don't run on forks) can run. Once that passes, I'll merge into main

k-capehart:
> @k-capehart Thanks for doing this work 🏆
> 
> I'm going to merge this into a separate feature branch so that all our integration tests (which don't run on forks) can run. Once that passes, I'll merge into main

No problem! And thank you for helping. I had fun working on this and learned a lot, so looking forward to being able to use it.

PR #932, opened 2024-01-29
feat: allow theme file to be configurable

mdonnalley:
Allow the location of the `theme.json` to be configured by the root CLI's package.json. This allows CLIs to ship a theme to their users

```json
"files": ["theme.json"],
"oclif": {
  "theme": "./theme.json"
}
```

If users have their own theme at `~/.confg/<CLI>/theme.json` then it will be merged in with the default theme, with the user theme taking precedence. 

@W-14527856@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -415,16 +415,12 @@ export class Config implements IConfig {
       await safeReadJson<Record<string, string>>(userThemeFile),
     ])
 
-    if (userTheme) {
-      return {
-        file: userThemeFile,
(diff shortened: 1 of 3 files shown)


mdonnalley:
@AllanOricil Thought you might be interested in this. We're finally getting to incorporating themes into `sf`. The plan is to have CLIs configure a default theme that ships with the CLI that users can override - either partially or entirely.

AllanOricil:
It is time for the RGB version running at 60fps!

shetzel (approved):

PR #976, opened 2024-02-27
Proposal: Add custom plugin loader

isaacroldan:
At Shopify we want to be able to dynamically load our own plugins for a very specific use case:

>If you are running a global CLI in a folder with a package.json, and that package.json
 has a dependency on something like `@shopify/*`, try to load that as a plugin.

That way plugins can be defined per project just as normal dependencies. (without adding an `oclif` entry in package.json)

--- 
We were doing this with `oclif v2` by subclassing `Config` and overriding the `loadCorePlugins()`. You can see how [here](https://github.com/Shopify/cli/blob/global-cli/packages/cli-kit/src/public/node/custom-oclif-loader.ts).

This is not possible in `oclif v3` because `loadCorePlugins` no longer exists and everything is abstracted in the PluginLoader.

This PR is a simple proposal to expose a method that would let us load custom plugins that are not defined in our package.json (or anywhere).

Please let me know if this is a valid approach or you'd prefer to do this in a different way.

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -375,6 +375,13 @@ export class Config implements IConfig {
     marker?.stop()
   }
 
+  public async loadCustomPlugins(root: string, plugins: string[]) {
+    const marker = Performance.mark(OCLIF_MARKER_OWNER, 'config.loadCustomPlugins')
+    await this.pluginLoader.loadCustomPlugins(root, plugins)
+    await this.loadPluginsAndCommands()
+    marker?.stop()
+  }
+
   async loadPluginsAndCommands(opts?: {force: boolean}): Promise<void> {
     const pluginsMarker = Performance.mark(OCLIF_MARKER_OWNER, 'config.loadAllPlugins')
     const {errors, plugins} = await this.pluginLoader.loadChildren({
--- a/src/config/plugin-loader.ts
+++ b/src/config/plugin-loader.ts
@@ -48,6 +48,10 @@ export default class PluginLoader {
     return {errors: this.errors, plugins: this.plugins}
   }
 
+  public async loadCustomPlugins(root: string, plugins: ({name?: string; root?: string; tag?: string; url?: string} | string)[],): Promise<void> {
+    await this.loadPlugins(root, 'custom', plugins)
+  }
+
   public async loadRoot(): Promise<IPlugin> {
     let rootPlugin: IPlugin
     if (this.pluginsProvided) {


isaacroldan:
I did sign the CLA (if I try again it says i've already signed). 
But I think I did the CLA with my @gmail address, while the commit has been signed with my shopify one :/

mdonnalley:
>I did sign the CLA (if I try again it says i've already signed).
But I think I did the CLA with my gmail address, while the commit has been signed with my shopify one :/

Don't worry about it - it's a pretty finicky process.

As far as the PR goes, I like the idea but I'm wondering if there's a better solution. Your current changes call `loadPluginsAndCommands` after loading the custom plugins which is going to slow down your CLI since it will cause all the plugins to be loaded again.

So I'm wondering if being able to specify wildcards in the plugins list would work for you. For example

```json
{
  "oclif": {
    "plugins": [
        "@oclif/plugin-plugins",
        "@shopify/*",
     ]
  }
}
```

Then `PluginLoader` will have method, `loadWildCardPlugins` that will match `@shopify/*` to any dependencies and load them, e.g.

```typescript
  private async loadWildCardPlugins(opts: LoadOpts): Promise<void> {
    if (opts.rootPlugin.pjson.oclif.plugins) {
      const {dependencies, oclif} = opts.rootPlugin.pjson
      const wildCardPlugins = oclif.plugins?.filter((p) => p.includes('*')) ?? []
      const plugins = Object.keys(dependencies ?? {}).filter((p) =>
        // Just a working example that works for things like `@shopify/*` or `@oclif/plugin-*`
        // we'd want to use a library like minimatch to fully support wildcards
        wildCardPlugins.some((w) => p.startsWith(w.slice(0, -1))),
      )
      await this.loadPlugins(opts.rootPlugin.root, 'core', plugins)
    }
  }
```

If don't want to have to specify anything in `oclif.plugins` then we could have a option on `Config` that tells it to load the wildcard plugins. Something like this:

```typescript
const config = new Config({ wildCardPlugins: ['@shopify/*'] })
await config.load()
```

Thoughts?

isaacroldan:
```ts
const config = new Config({ pluginAdditions: ['@shopify/*'] })
await config.load(
```

I like this idea!

But could we also pass a path for where to look for those plugins? our CLI has the `--path` flag that basically let's you run it from everywhere but point to that path. Like `shopify app deploy --path /path/to/my/app`

So we would like to be able to load any `@shopify/*` plugin installed at `/path/to/my/app`. Does that make sense?
That's why in this PR i'm passing the `root: string` argument.

We do that in our current implementation for v2 [here](https://github.com/Shopify/cli/blob/global-cli/packages/cli-kit/src/public/node/custom-oclif-loader.ts#L28-L31) and [here](https://github.com/Shopify/cli/blob/global-cli/packages/cli-kit/src/public/node/custom-oclif-loader.ts#L143-L152) we get the path value

Thanks for your support @mdonnalley 🙏

Issue #944, opened 2024-02-08
ux.Table: Incorrect parsing of double quotes when exporting to CSV


julesbonnard:
**Describe the bug**
The CSV output of ux.Table only escapes the first double quote of a string value, resulting in a bad CSV format when exporting string values with multiple double quotes followed by a comma.

**To Reproduce**
Try exporting with --csv flag a table containing the following string: `"That film", a great movie`.
You get this wrong result: `"""That film", a great movie"`

**Expected behavior**
According to CSV spec `If double-quotes are used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote.`, the correct export should be `"""That film"", a great movie"`

**Additional context**
Problem seems to come from this line : https://github.com/oclif/core/blob/6b9af08e56990ceede201087069e8f60dde125af/src/cli-ux/styled/table.ts#L153, where replaceAll should be used instead of replace.

Also, it would be great to only escape values containing characters like double quotes, commas and line breaks, and not the entire line

Thanks for your help

mdonnalley:
@julesbonnard Sounds like you've around got your head wrapped around a solution - any chance you'd be willing to submit a PR?

julesbonnard:
Sure, I'm not used to but I'll try.

Le jeu. 22 févr. 2024, 22:41, Mike Donnalley ***@***.***> a
écrit :

mdonnalley:
Closed by #980

PR #977, opened 2024-02-28
Fix: correctly escapes double quotes when converting ux.table to csv

julesbonnard:
modify replace with replaceAll as mentioned in this issue : https://github.com/oclif/core/issues/944#issue-2126189996


mdonnalley (changes requested):
Code change looks good but can you add a test case here: https://github.com/oclif/core/blob/main/test/cli-ux/styled/table.test.ts#L150

Thanks!

julesbonnard:
Thanks for your review, I've just added a new test case.

I have one more question for the future: should the check for escaping quotes be done line by line (as it is now) or value by value? The latter would produce slightly cleaner CSVs, but at some performance cost.

mdonnalley (approved):

PR #1046, opened 2024-04-08
fix: allowStdin with no input

k-capehart:
- do not throw an error if a flag with allowStdin='only' is immediately followed by another flag
- include a new test

closes https://github.com/forcedotcom/cli/issues/2811


mdonnalley (approved):

mdonnalley:
Thanks @k-capehart for the fix 🏆 

I just merged this into a separate branch so all the integration tests can run. Once those pass, I'll merge that to main

mdonnalley:
@k-capehart Thanks for the reminder - I just made a work item on our internal tracker for it. Should be addressed fairly soon.

PR #1056, opened 2024-04-15
feat: slimmer ux

mdonnalley:
### Slimmer `ux`
As described [here](https://github.com/oclif/core/discussions/999), we're removing most of the methods in the `ux` module. We're simply unable to adequately support the feature set that `ux` offers and think that most people would benefit from using dedicated libraries that are better supported.

We are, however, keeping some of the functionality. The new `ux` module will contain the following:

#### Unchanged

- `colorize`
- `error`
- `exit`
- `action` - will be unchanged from previous version except that the spinner color will be configurable using [themes](https://oclif.io/docs/themes).
- `warn`

#### Renamed

- `stdout` - rename of `ux.log`
- `stderr` - rename of `ux.logToStderr`

#### New

- `colorizeJson` - Apply color theme to arbitrary JSON.

#### What you'll need to do

You will need to replace everything that `ux` was doing with dedicated libraries. Here are a few suggestions:

- For prompts: [inquirer](https://www.npmjs.com/package/inquirer)
- For progress bars: [cli-progress](https://www.npmjs.com/package/cli-progress)
- For hyperlinks: [hyperlink](https://www.npmjs.com/package/hyperlink)
- For tables: [tty-table](https://www.npmjs.com/package/tty-table), [cliui](https://www.npmjs.com/package/cliui)
- For trees: [object-treeify](https://www.npmjs.com/package/object-treeify)
- For notifications: [node-notifier](https://www.npmjs.com/package/node-notifier)
- For links: [terminal-link](https://www.npmjs.com/package/terminal-link)
- For rendering react components: [ink](https://www.npmjs.com/package/ink)

### No fancy-tests
- Removes all uses of `@oclif/test` and `fancy-test`
- Adds new test utilities in `test/test.ts`, which will soon become the next major of `@oclif/test`


src/config/config.ts
```diff
@@ -333,15 +334,17 @@ export class Config implements IConfig {
 
     this.npmRegistry = this.scopedEnvVar('NPM_REGISTRY') || this.pjson.oclif.npmRegistry
 
-    if (!this.scopedEnvVarTrue('DISABLE_THEME')) {
-      const {theme} = await this.loadThemes()
-      this.theme = theme
+    this.theme = await this.loadTheme()
+
+    const s3 = this.pjson.oclif.update?.s3 ?? {
```
mshanemc, line 339: I don't like this style (declaring a const object with placeholders).

could there be a function whose single responsibility is to build the `s3` object, or the `updateConfig` object?

src/config/config.ts
```diff
@@ -97,11 +97,13 @@ export class Config implements IConfig {
   public shell!: string
   public theme?: Theme
   public topicSeparator: ' ' | ':' = ':'
+  public updateConfig!: NonNullable<PJSON.CLI['oclif']['update']>
```
mshanemc, line 100: if we're doing breaking changes, maybe we could make the types honest?  Like, these `!` are kinda true as long as you've called the static `load` method?  

And TS can't help developers avoid runtime undefineds if we do stuff like that.

noodle on it

src/config/config.ts
mshanemc: and if the property is optional, we could type it that way rather than using empty string

src/errors/warn.ts
```diff
@@ -1,10 +1,19 @@
...
 import {CLIError, addOclifExitCode} from './errors/cli'
 import prettyPrint from './errors/pretty-print'
 
-export function warn(input: Error | string): void {
+const WARNINGS = new Set<Error | string>()
+
+/**
+ * Prints a pretty warning message to stderr.
+ */
+export function warn(input: Error | string, options?: {ignoreDuplicates: boolean}): void {
```
mshanemc, line 12: maybe jsdoc the ignoreDuplicates option?

src/help/command.ts
```diff
@@ -377,9 +376,9 @@ export class CommandHelp extends HelpFormatter {
   }
 
   private isCommand(example: string): boolean {
-    return stripAnsi(this.formatIfCommand(example)).startsWith(
-      `${colorize(this.config?.theme?.dollarSign, '$')} ${this.config.bin}`,
-    )
+    return ansis
```
mshanemc, line 379: ooh, that's nice

src/ux/theme.ts
```diff
@@ -0,0 +1,43 @@
...
+  return text
+}
+
+export function parseTheme(theme: Record<string, string>): Theme {
+  return Object.fromEntries(
+    Object.entries(theme)
+      .map(([key, value]) => {
+        if (typeof value === 'string') return [key, isValid(value)]
+        return [key, parseTheme(value)]
+      })
```
mshanemc, line 36:
```suggestion
      .map(([key, value]) => [key, typeof value === 'string' ? isValid(value) :  parseTheme(value)]})
```

mshanemc (approved):

PR #1067, opened 2024-04-26
feat: support rc files

mdonnalley:
Use [cosmiconfig](https://www.npmjs.com/package/cosmiconfig) to support rc files, which will allow consumers to use an rc file to configure oclif instead of using the package.json.

Important to note that oclif will still be dependent on the package.json to get the `name`, `version`, and `dependencies`

The following files are supported:
```
.oclifrc
.oclifrc.json
.oclifrc.yaml
.oclifrc.yml
.oclifrc.js
.oclifrc.ts
.oclifrc.mjs
.oclifrc.cjs
oclif.config.js
oclif.config.ts
oclif.config.mjs
oclif.config.cjs
```

@W-15470586@

### DIFF
--- a/src/util/read-pjson.ts
+++ b/src/util/read-pjson.ts
@@ -8,7 +8,7 @@ import {readJson} from './fs'
 const debug = makeDebug('read-pjson')
 
 /**
- * Read the package.json file from a given path and merge with the config found by cosmiconfig.
+ * Read the package.json file from a given path and add the oclif config (found by cosmiconfig) if it exists.
  *
  * We can assume that the package.json file exists because the plugin root has already been loaded at this point.
  */
@@ -59,6 +59,6 @@ export async function readPjson(path: string): Promise<PJSON.Plugin> {
   debug(`found oclif config for ${path}: %O`, result)
   return {
     ...pjson,
-    oclif: result?.config ?? pjson.oclif ?? {},
+    oclif: result?.config ?? {},
   }
 }
--- a/src/util/read-pjson.ts
+++ b/src/util/read-pjson.ts
@@ -1,0 +1,64 @@
+import {cosmiconfig} from 'cosmiconfig'
+import {join} from 'node:path'
+
+import {PJSON} from '../interfaces'
+import {makeDebug} from '../logger'
+import {readJson} from './fs'
+
+const debug = makeDebug('read-pjson')
+
+/**
+ * Read the package.json file from a given path and merge with the config found by cosmiconfig.
+ *
+ * We can assume that the package.json file exists because the plugin root has already been loaded at this point.
+ */
+export async function readPjson(path: string): Promise<PJSON.Plugin> {
+  const pjsonPath = join(path, 'package.json')
+  if (process.env.OCLIF_DISABLE_RC) {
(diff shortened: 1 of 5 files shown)


src/util/read-pjson.ts
```diff
@@ -0,0 +1,64 @@
...
+      '.oclifrc.yaml',
+      '.oclifrc.yml',
+      '.oclifrc.js',
+      '.oclifrc.ts',
+      '.oclifrc.mjs',
+      '.oclifrc.cjs',
+      'oclif.config.js',
+      'oclif.config.ts',
+      'oclif.config.mjs',
+      'oclif.config.cjs',
```
iowillhoit, line 49: Is there a good use-case to support so many filetypes here? I feel like just supporting `.oclifrc.json` would be plenty. Then we could get rid off the `cosmicconfig` package.

src/util/read-pjson.ts
```diff
@@ -0,0 +1,64 @@
...
+import {join} from 'node:path'
+
+import {PJSON} from '../interfaces'
+import {makeDebug} from '../logger'
+import {readJson} from './fs'
+
+const debug = makeDebug('read-pjson')
+
+/**
+ * Read the package.json file from a given path and merge with the config found by cosmiconfig.
```
iowillhoit, line 11: This is slightly misleading. It makes it sound like we are merging `pjson.oclif` config with with additional config discovered by `cosmicconfig`.

mdonnalley: Yeah I think "merge" as in, "put an `oclif` prop in the package.json"... I'll clarify it

src/util/read-pjson.ts
mdonnalley: `js` and `ts` files offer several benefits:

- comments
- dynamically set props
- use types

iowillhoit:
In my initial attempt to use a stand-alone config file, i copied the entire `"oclif: {}"` object out of my `package.json` and it would not work. I then looked at the code and saw that i needed to remove the `"oclif"` key and just use the values. We should make sure an example is documented.

iowillhoit:
## QA NOTES

QA looks good

- 🟢 `package.json` config is used if it exists
- 🟢 Works with `.oclifrc`
- 🟢 Works with `.oclifrc.json`
- 🟢 .oclifrc.mjs
    - `export default {`
- 🟢 oclif.config.mjs
    - `export default {`
- 🟢 .oclifrc.ts
    - `export default {`
- 🟢 oclif.config.ts
    - `export default {`
- 🟢 .oclifrc.cjs
    - `module.exports = {`
- 🟢 oclif.config.cjs
    - `module.exports = {`
- 🟢 oclif.config.js
    - `export default {`
    - `module.exports = {`
- 🟢 .oclifrc.js
    - `export default {`
    - `module.exports = {`
- 🟢 .oclifrc.yaml
- 🟢 .oclifrc.yml
    - For yaml:
```
bin: oclif-cli-rc
dirname: oclif-cli-rc
commands: ./dist/commands
plugins:
  - "@oclif/plugin-help"
  - "@oclif/plugin-plugins"
topicSeparator: " "
topics:
  hello:
    description: "Say hello to the world and others"
```

iowillhoit (approved):

PR #1071, opened 2024-05-04
fix: new "dont check npm" option for plugin-update 

gregsadetsky:
this PR adds a type in order for [the other PR](https://github.com/oclif/plugin-update/pull/807) on oclif/plugin-update to work

### DIFF
--- a/src/interfaces/pjson.ts
+++ b/src/interfaces/pjson.ts
@@ -149,7 +149,7 @@ export namespace PJSON {
           debounce?: number
           rollout?: number
         }
-        disableNpmLookup: boolean
+        disableNpmLookup?: boolean
         node: {
           targets?: string[]
           version?: string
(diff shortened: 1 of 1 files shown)


src/interfaces/pjson.ts
```diff
@@ -149,6 +149,7 @@ export namespace PJSON {
           debounce?: number
           rollout?: number
         }
+        disableNpmLookup: boolean
```
mdonnalley, line 152: Can you make this optional? `disableNpmLookup?: boolean`

mdonnalley (changes requested):

src/interfaces/pjson.ts
gregsadetsky: Of course, my bad! Done

mdonnalley (approved):

Issue #1088, opened 2024-05-30
Option to customize "<value>" in docopts help usage strings


kemley76:
**Is your feature request related to a problem? Please describe.**
I am working on a CLI that currently has a lot of manually written usage strings (seen in help menus) for commands. I want to remove them to allow Oclif's docopts functionality to autogenerate them, but this would force the removal of helpful information only the manual formatting could contain.

Example: 
```
USAGE
    $ mycli add -i <input-files>...
```
would be autogenerated as
```
USAGE
    $ mycli add -i <value>
```

This "`<value>`" seems to be a hardcoded value, even when changing the flag's `helpValue` property. It also does not include the `...` to indicate multiple inputs are allowed. The [docopts site](http://docopt.org/) uses these notations quite a bit it seems. 

**Describe the solution you'd like**
I would like an optional flag property (separate from the current `helpValue`) that can control this value, which could default to `'<value>'`. It would be used in autogenerating the docopts for the usage strings in the help menus. Perhaps this sort of feature could be extended to positional arguments, but that is not really my main concern. 

I'd appreciate any help or guidance you could provide on this issue.

mdonnalley:
>I would like an optional flag property (separate from the current helpValue) that can control this value

@kemley76 what's the rationale behind having a separate property? I'd imagine most people who are setting `helpValue` would want to same to be used in the usage

kemley76:
I guess there is no real reason to have it be a separate property. I figured that would allow for more flexibility, but I agree, I don't see a reason to have them different. For what I wanted to do with it, extending the current `helpValue` is all I would need.

mdonnalley:
Got it. We can always add a separate property later if people need it. Do you want to update your PR to use `helpValue`? If so, I'll review and test it out this week

kemley76:
My PR is now updated.

PR #1091, opened 2024-05-31
feat: added option to specify example flag value in docopts

kemley76:
Feature add for #1088 
- Made changes to the docopts formatter
- Wrote a test for the docopts changes

There is the option to include multiple example flags such that the usage string can format to something like 
`mycli -i <input-json>... <input-xml>...`


mdonnalley (approved):

mdonnalley:
Thanks @kemley76 for the contribution 🏆

PR #1134, opened 2024-07-08
fix: correctly identify powershell

mdonnalley:
Correctly distinguish between powershell and cmd.exe

@W-11446819@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -663,6 +663,10 @@ export class Config implements IConfig {
     const SHELL = process.env.SHELL ?? osUserInfo().shell?.split(sep)?.pop()
     if (SHELL) {
       shellPath = SHELL.split('/')
+    } else if (this.windows && process.title.toLowerCase().includes('powershell')) {
+      shellPath = ['pwsh']
+    } else if (this.windows && process.title.toLowerCase().includes('command prompt')) {
+      shellPath = ['cmd.exe']
     } else if (this.windows && COMSPEC) {
       shellPath = COMSPEC.split(/\\|\//)
     } else {
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -659,12 +659,17 @@ export class Config implements IConfig {
 
   protected _shell(): string {
     let shellPath
-    const {COMSPEC} = process.env
+    const {COMSPEC, PSModulePath} = process.env
     const SHELL = process.env.SHELL ?? osUserInfo().shell?.split(sep)?.pop()
     if (SHELL) {
       shellPath = SHELL.split('/')
+    } else if (!this.windows && PSModulePath) {
+      // PSModulePath is always set in powershell, regardless of the OS
+      // We can't use this on windows though since both powershell and cmd.exe set it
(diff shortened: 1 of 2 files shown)


src/config/config.ts
```diff
@@ -663,6 +663,10 @@ export class Config implements IConfig {
     const SHELL = process.env.SHELL ?? osUserInfo().shell?.split(sep)?.pop()
     if (SHELL) {
       shellPath = SHELL.split('/')
+    } else if (this.windows && process.title.toLowerCase().includes('powershell')) {
+      shellPath = ['pwsh']
```
cristiand391, line 667: the `process.title` check works on windows but powershell can be used on mac/linux where the title is the node bin path.

Screenshot 2024-07-08 at 3 02 59 PM

what if we look for a powershell-specific env var to detect it on any OS?
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.4

there are a few windows-specific but `PSModulePath` is set on all OSes by default:
> The $env:PSModulePath environment variable contains a list of folder locations that are searched to find modules and resources. On Windows, the list of folder locations is separated by the semi-colon (;) character. On non-Windows platforms, the colon (:) separates the folder locations in the environment variable.

mac:
Screenshot 2024-07-08 at 3 07 20 PM

win:
Screenshot 2024-07-08 at 3 09 01 PM

cristiand391 (changes requested):
left one suggestion to improve detection on linux/mac

cristiand391 (approved):

src/config/config.ts
cristiand391: follow up:
`PSModule` is also available in `cmd` ... will be using both checks.

cristiand391:
**QA notes**:

setup:

linked this branch of oclif into plugin-autocomplete 
win: patched oclif/core in sf install folder

general:
✅ `SHELL` still takes precedence. This is usually set at the system-level, can be modified in terminals/multiplexers when needed

mac:
🟡 powershell detected:
because `SHELL` is set by default,  the only way to make oclif detect powershell correctly is to make your term/multiplexer unset it/set to `SHELL=powershell`. So, the current behavior of detecting powershell on mac/linux doesn't change.

win:
✅ powershell detected
`SHELL` isn't set and `os.userInfo().shell` is always null:
https://nodejs.org/api/os.html#osuserinfooptions
> Returns information about the currently effective user. On POSIX platforms, this is typically a subset of the password file. The returned object includes the username, uid, gid, shell, and homedir. On Windows, the uid and gid fields are -1, and shell is null.

so oclif does the `process.title` check.
I can run `sf autocomplete` alone instead of `sf autocomplete powershell` and get valid pwsh instructions.

Screenshot 2024-07-08 at 4 11 45 PM

here I ran `sf autocomplete --dev-debug` on windows's cmd to verify `config.shell` value is still cmd.exe (this is also set by the process.title check b/c the `COMPSEC` one has lower precedence.
Screenshot 2024-07-08 at 4 12 35 PM

cristiand391 (dismissed):

PR #1148, opened 2024-07-26
fix: ignore escaped delimiters

mdonnalley:
Ignore escaped delimiters when splitting input string for flags with `delimiter` and `multiple`

**Before**
```
my-cli --flag one,two,three` => {flag: ['one', 'two', 'three']}
```

**After**
```
my-cli --flag one\,two,three` => {flag: ['one,two', 'three']}
```

This is useful when you have flag values that contain the delimiter but shouldn't be split.

_Depending on you shell, you might need to double escape to get the desired effect._

@W-16343745@
https://github.com/forcedotcom/cli/issues/2928

### DIFF
--- a/src/parser/parse.ts
+++ b/src/parser/parse.ts
@@ -412,17 +412,27 @@ export class Parser<
 
         // multiple with custom delimiter
         if (fws.inputFlag.flag.type === 'option' && fws.inputFlag.flag.delimiter && fws.inputFlag.flag.multiple) {
+          // regex that will identify unescaped delimiters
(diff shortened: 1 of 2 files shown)


mshanemc:
QA: threw extra UT at it, all ✅ 

the breaking scenario would be "I'm already using `multiple` and `delimiter: ','` AND I have the escape character like `\,` or `\\,` being used by my command and that's currently being interpreted as a delimiter and now it's about to not be.

seems pretty low-probability.

mshanemc (approved):

PR #1184, opened 2024-09-03
fix: handle large throughput writes

### LINKED ISSUE #1177: process.stdout buffering issue
I am experiencing and issue where lines of the help output are intermitently missed. It appears to be related to buffering/drain on the writes to `process.stdout`.

Specifically, the `stdout`/`stderr` calls to `write()` are not checking the return code, and not waiting for the `drain` event if/when required.

See https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback

https://github.com/oclif/core/blob/f38f8270bafa04cdcb8e6d193b808e099e205864/src/ux/write.ts#L3-L25

I have replaced all calls with synchronous writes as follows and this appears to resolve the issue:

```ts
import {writeSync} from 'node:fs'

...
writeSync(process.stdout.fd, format(...str, ...args) + '\n')
```

Thoughts?

Has anyone else experienced this?

mdonnalley:
Use `console.log` and `console.error` instead of `process.stdout.write` and `process.stderr.write`, to avoid losing information from back pressure on the streams

To replicate:

```js
// write.js

import {stdout} from './src/ux/write.js'
const lines = Array.from(
  {length: 100},
  (_, i) =>
    `Line ${i} Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer convallis fringilla sollicitudin. Nunc scelerisque neque non ipsum accumsan commodo. In et porttitor eros, ut vestibulum magna. Morbi felis diam, pharetra eu dui non, sollicitudin feugiat nisi. Aliquam cursus malesuada risus, vel luctus leo ornare sed. Morbi condimentum odio id ex facilisis bibendum. Nullam consectetur consectetur viverra. Donec nec ante dui. Integer lacinia facilisis urna vitae feugiat.`,
)

for (const line of lines) {
  stdout(line)
}
```

```
bun write.js
```

**NOTE** that the issue can only be replicated with bun - not node

Fixes #1177 

@W-16645420@

### DIFF
--- a/src/ux/write.ts
+++ b/src/ux/write.ts
@@ -1,25 +1,24 @@
 import {format} from 'node:util'
-
 export const stdout = (str?: string | string[] | undefined, ...args: string[]): void => {
   if (!str && args) {
-    process.stdout.write(format(...args) + '\n')
+    console.log(format(...args))
   } else if (!str) {
-    process.stdout.write('\n')
+    console.log()
   } else if (typeof str === 'string') {
-    process.stdout.write((str && format(str, ...args)) + '\n')
+    console.log(format(str, ...args))
   } else {
-    process.stdout.write(format(...str, ...args) + '\n')
+    console.log(format(...str, ...args))
   }
 }
 
 export const stderr = (str?: string | string[] | undefined, ...args: string[]): void => {
   if (!str && args) {
-    process.stderr.write(format(...args) + '\n')
+    console.error(format(...args))
   } else if (!str) {
-    process.stderr.write('\n')
+    console.error()
   } else if (typeof str === 'string') {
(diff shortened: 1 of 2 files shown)


src/ux/write.ts
danielsharvey, line 1: LGTM. I think it makes good sense to use console.log/console.error.

test/ux/write.test.ts
danielsharvey, line 1: Does it make sense to put a comment with the tests to provide context as to the motivation?
i.e. previous use of process.stdout/process.stderr had backpressure issues on Bun

shetzel (approved):

shetzel:
Using the repro script I saw skipped lines without the changes.  With the changes no lines were skipped.

PR #1229, opened 2024-10-22
fix: fix aliases in config.commandIDs missing the default topic separator

myarmolinsky:
This fixes the following issue/example:
- `cli add topic3` is an alias for `cli set topic3`
- `cli add` is an alias for `cli set`
- Try to pipe content to the alias command: `cat foo | cli add topic3`
- The content of `foo` *should* be sent to `cli add topic3`, but it is instead sent to `cli add`
- In contrary, `cat foo | cli set topic3` works fine: the content of `foo` is sent to `cli set topic3` rather than `cli set`

Please let me know if I am missing some context, or if there is a better spot to apply this change. I was looking for a spot to apply it before `Config.load` is called (or at least earlier on than when `loadPluginsAndCommands` is called in `Config.load`) but I couldn't find an appropriate spot

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -17,6 +17,7 @@ import {OCLIF_MARKER_OWNER, Performance} from '../performance'
 import {settings} from '../settings'
 import {determinePriority} from '../util/determine-priority'
 import {safeReadJson} from '../util/fs'
+import {toStandardizedId} from '../util/ids'
 import {getHomeDir, getPlatform} from '../util/os'
 import {compact, isProd} from '../util/util'
 import {ux} from '../ux'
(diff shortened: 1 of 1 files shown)


thgreasi (approved):

mdonnalley:
@myarmolinsky Thanks for taking the time to make the PR! 

Quick question: how are you defining the aliases? I found that I'm only able to replicate the issue if I set the aliases like this: `aliases = ['add topic3']` (space separated). If set it to `aliases = ['add:topic3']` (colon separated) - it works

We never intended for that `aliases` array to be space separated - although I can't think of a good reason not to support it

src/config/config.ts
```diff
@@ -780,23 +780,27 @@ export class Config implements IConfig {
       }
 
       const handleAlias = (alias: string, hidden = false) => {
-        if (this._commands.has(alias)) {
+        const aliasWithDefaultTopicSeparator = alias.replaceAll(' ', ':')
```
mdonnalley, line 783: We have a util function for whenever we need to do this:

```ts
import { toStandardizedId } from '../util/ids'
const aliasWithDefaultTopicSeparator = toStandardizedId(alias)
```

mdonnalley (changes requested):
Looks good!

Just need one change then I'll merge into a separate feature branch so that all the integration tests can run. Once those pass, I'll merge to main

myarmolinsky:
Thank you for always being so quick to review!

I am indeed setting the aliases as `aliases = ['add topic3']`. We use ` ` as a topic separator instead of `:` in the [balena CLI](https://github.com/balena-io/balena-cli). In this particular case we had a command `balena key add` which I created an alias for as I intend to deprecate the old one. The new alias is `balena ssh-key add`. Given that the old command had the space when we would use it (as well as in any help/warning output users would receive, including the deprecated command warning as I had [committed previously](https://github.com/oclif/core/commit/fb6ace237c5f6b3f7707edb7ba93e00aeee9b412)), my instinct was to add it to the `aliases` array with the ` ` rather than with a `:`.

I have just tested `aliases - [add:topic3]` instead of with the space and I see that it indeed still appears to print the correct expected output help/warnings, which is great to see as well. In general I suppose the ideal UX for us would be for everything to respect/expect the topic separator of choice (hence this PR and the previous one).

src/config/config.ts
myarmolinsky: 🤦 I should have noticed this, my bad

PR #1331, opened 2025-03-05
feat: export validate from the Parse module

jpshack-at-palomar:
The oclif project I am working on includes a features that allow one to orchestrate the execution of a pipeline of Oclif commands. Since it supplies JSON to each command with args and flags it is wasteful to put the json back into a command line and then parse the flags and args out again. That said, we do want to validate the args and flags but now that Oclif better controls is exports, we do not have access to the needed validate function. This PR restores access to the validate call so that we do not have have to copy and reproduce all of that code in our project. We have been working off a private fork for this one change but are nearing release and would like to make things right ahead of that.

Thanks for your consideration!

### DIFF
--- a/src/parser/index.ts
+++ b/src/parser/index.ts
@@ -6,6 +6,7 @@ import {validate} from './validate'
 export type {ArgInput, FlagInput, Input, OutputArgs, OutputFlags, ParserOutput} from '../interfaces/parser'
 
 export {flagUsages} from './help'
+export {validate} from './validate'
 
 export async function parse<
   TFlags extends OutputFlags<any>,


cristiand391:
> Since it supplies JSON to each command with args and flags it is wasteful to put the json back into a command line and then parse the flags and args out again. That said, we do want to validate the args and flags 

`validate` requires a parser call already, how are you skipping parsing currently?
https://github.com/oclif/core/blob/95afb53de93218799dd5d8195794a260acc9b64a/src/parser/index.ts#L23C2-L25C34

For some specific code/types/interfaces we purposely stopped exporting them (easier to make changes in  the parser, public types without public breaking changes), not sure if `validate` was in this case but it's tightly coupled to the parser.

cristiand391 (approved):

cristiand391:
thanks for the context @jpshack-at-palomar 

I think it's ok, we already expose the interface by exporting the parser so even if we didn't export `validate` we would still need a major release for breaking changes there.

cristiand391:
just 1 flakyNUT failure from plugin-data, safe to ignore:
```
     Error: Deleting org test-4xnjftenuxpx@example.com failed due to: Error (1): ConcurrentRequests (Concurrent API Requests) Limit exceeded.
```

PR #1412, opened 2025-07-02
fix: do not list hidden arguments in usage

rexxars:
This addresses an issue where the hidden flag for arguments was not respected in usage output.

Theoretically I wonder if we should also be checking/throwing if trying to define commands that have hidden arguments _before_ non-hidden arguments. Eg consider:

```
args: {
  foo: Args.string({hidden: true}),
  bar: Args.string({})
}
```
… there's no way you can specify `bar` but not `foo`, so this doesn't really make sense in my book? But… a bit of scope creep, so will leave that for a later PR/someone else to figure out :)

### DIFF
--- a/src/help/docopts.ts
+++ b/src/help/docopts.ts
@@ -101,9 +101,11 @@ export class DocOpts {
       // If strict is false, add ellipsis to indicate that the argument takes multiple values
       const suffix = this.cmd.strict === false ? '...' : ''
       const a =
-        Object.values(ensureArgObject(this.cmd.args)).map((arg) =>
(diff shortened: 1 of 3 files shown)


mdonnalley (approved):

mdonnalley:
Thanks for the PR! Changes look good to me. I'll merge into a separate branch so integration tests can run - once that passes I'll merge to main

>Theoretically I wonder if we should also be checking/throwing if trying to define commands that have hidden arguments before non-hidden arguments

My philosophy typically has been to allow greater flexibility, even if it allows developers to create poorly designed commands. That being said, I'd accept a PR to prevent that scenario from happening

PR #1435, opened 2025-07-23
fix(W-19118139): standardize aliases

mdonnalley:
Fixes https://github.com/oclif/plugin-autocomplete/issues/1004

@W-19118139@


cristiand391:
**QA notes**:

✅ colon-separated aliases still work
✅ mixed aliases in the same command (colon and space separated) are supported

PR #1453, opened 2025-08-27
fix: indicate whether an argument is optional in the help command

myarmolinsky:
Makes the help command `args()` function distinguish optional arguments


myarmolinsky:
Thanks @mdonnalley , I have dropped the first commit and updated the second to only apply `[]`, not `<>`

mdonnalley (approved):

PR #1461, opened 2025-09-08
chore: remove duplicate "one-way exclusive optional field" docopts test

thgreasi:
The input & output of the removed test is exactly the same as the one right above
https://github.com/oclif/core/blob/v1.16.4/test/help/docopts.test.ts#L148-L182


thgreasi:
Any update on this?

PR #1487, opened 2025-10-19
feat: support the 'combinable' & 'only' flag relationships

thgreasi:
This new flag relationship aims to ease defining that a specific flag is ONLY compatible with just a few other ones.
Currently users would have to enumerate all other incompatible flags in the exclusive relationship, which leads to some repetition, doesn't capture exactly the original intention, and could lead to oversights if more flags are added to that command later. I realized this use case while working on a balena-cli PR, where we had to define that
 --config is allowed to be combined with the --yes flag and is incompatible with all other flags.
The workaround that we ended up having to use, was to use an IIFE and define the compatible flags in a separate array and reference that in the exclusive property. https://github.com/balena-io/balena-cli/pull/2979/commits/692e70d51883204217fb0d5bdebb976807c94a85

I'm sure I have missed some pieces like adding proper docs, but thought it would be better to first check whether there would be interest in extending the flag relationships. For sure plz suggest any better name that you can think.


mdonnalley (approved):

mdonnalley:
@thgreasi Just tested this out and it works great! Thanks for making the change

I'll merge this into a separate feature branch so that integration tests can run. Once those pass, I'll merge to main

PR #1568, opened 2026-03-30
feat: add `Args.option()` for typed arg options

rexxars:
Adds a new `Args.option()` method mirroring `Flags.option()`, which provides type-safe option usage:

```ts
class Deploy extends Command {
  static args = {
    env: Args.option({
      options: ['production', 'staging', 'development'] as const
    })()
  }

  async run() {
    const {args} = await this.parse(Deploy)
    // `args.env` typed as `'production' | 'staging' | 'development' | undefined`
  }
}
```

Setting it as `required` removes the `undefined` from the union.

### DIFF
--- a/src/args.ts
+++ b/src/args.ts
@@ -99,7 +99,7 @@ type ReadonlyElementOf<T extends ReadonlyArray<unknown>> = T[number]
  * }
  */
 export function option<T extends readonly string[]>(
-  defaults: Partial<Arg<ReadonlyElementOf<T>>> & {options: T},
+  defaults: Omit<Partial<Arg<ReadonlyElementOf<T>>>, 'default' | 'multiple' | 'required'> & {options: T},
 ): ArgDefinition<(typeof defaults.options)[number]> {
   return (options: any = {}) => ({
     parse: async (i: string, _context: Command, _opts: Record<string, unknown>) => i,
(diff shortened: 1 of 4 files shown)


rexxars:
@jfeingold35 Any thoughts on this?

jfeingold35:
@rexxars , sorry for the delay! I'll get this reviewed shortly and talk it over with the team.

jfeingold35:
@rexxars , the code looks clean to me. We'll do some QAing and then we can get this approved and merged. However, it would be ideal if you could have a few runtime tests, e.g.,
- can it parse one arg
- Does it play well with `multiple`
- Does it reject invalid args
etc
Also, [the documentation](https://oclif.io/docs/args) would need to be updated as well.

rexxars:
Will have a look. I couldn't find any documentation on `Flags.option()`, so wasn't sure if this was meant to be "public" or not.

rexxars:
Rebased and added more tests. As for docs; do you want me to also document `Flags.option()`? IMHO the API for these are both a bit odd, but I'm not sure if you'd accept the amount of type overloading/changes needed to make it work if you specify the `options` _option_

jfeingold35:
@rexxars , I did some QA and it all seems to be working correctly, and the new tests all look good. I'll figure out why we're seeing the test failures currently, and let you know if you need to do anything to resolve them.
In the meantime, I would prefer we waited to firmly document the `Args.option()` or `Flags.option()`, just for the moment. We very well might add that documentation in the near future, but I'd like to take it one thing at a time.

jfeingold35 (approved):

PR #1580, opened 2026-04-17
feat: add command-level constraint framework @W-20534184@

### LINKED ISSUE #1507: New flag Constraints API
obs: I had this insight because I never really liked the way constraints are configured in oclif. The ticket with the argumentation to defend the introduction of this new feature was written by AI, but the API was my idea.

**Is your feature request related to a problem? Please describe.**

Oclif’s current `flags.relationships` feature forces developers to encode validation rules inside individual flags, which fragments business logic across multiple definitions and increases cognitive load. 

This results in:
- duplicated logic (exclusive has to be added in multiple places)
- impossible constraints (e.g., true XOR / “one-of-required”)
- scattered dependency rules (dependsOn, implies)
- maintenance overhead (one new flag often forces rewriting several others)
- unintuitive validation behavior when multiple rules interact

For anything beyond trivial commands, developers end up manually re-implementing logic inside run(), defeating the purpose of declarative flag relationships.

This creates real friction and makes flag-heavy commands difficult to maintain at scale.

**Describe the solution you'd like**

Introduce a centralized `defineConstraints` DSL that declares all flag relationships in a single, declarative structure. This shifts the mental model from “each flag owns its own rules” to “the command defines the rules; flags define types.”

This focuses the developer on business logic instead of scattering rules across the flag map.

----

### Example 1 — Simple and readable

```ts
import { xor, oneOf, require, forbid, defineConstraints } from '@oclif/constraints';

static constraints = defineConstraints([
  xor('upload', 'download').required(),
  oneOf('json', 'xml').required(),
  require('token').when('login'),
  forbid('debug').when('production'),
]);
```
This DSL compiles into a clear, predictable validation schema without the developer needing to understand complex internals.

### Example 2 — Mixed DSL + schema hybrid (supported)

```ts
static constraints = defineConstraints([
  xor('a', 'b').required(),
  {
    allOf: [['region', 'accountId']]
  }
]);
```

Both syntaxes are supported. DSL is ergonomic; schema is serializable = DSL turns into a structure json that can be serialized.

**Describe alternatives you've considered**

1. Current “relationships” API
- exclusive can prevent two flags from being used together but cannot require that one of them is present
- dependsOn cannot express grouped conditions
- implies cannot express mutual requirements or XOR
- all rules must be duplicated across flags
leads to contradictory or incomplete rule sets

2. Manual validation in run()
- extremely verbose
- not reusable
- bypasses Oclif’s declarative philosophy
error messages inconsistent across commands

3. Adding more relationships to flags
- makes flags bloated
- increases cognitive load
- does not fix the fundamental issue: rules belong to commands, not flags

**Additional context**

### Centralization eliminates cognitive load

Today, developers must bounce between 5–10 flags to reconstruct what the command actually expects. With a constraints DSL:

- logic lives in one place
- business rules are explicit
- onboarding new maintainers is trivial
- changes require updating only one file and one structure

### The DSL is forward-thinking
It opens the door for:

- auto-generated docs
- static analysis
- validation previews
- better error messages
- cross-command rule sharing

### Flags become clean

No more:

```ts
required: true
exclusive: ['x', 'y']
dependsOn: ['token']
implies: ['foo']
```

Flags declare shape only:
```ts
Flags.string()
Flags.boolean()
Flags.integer()
```

All logic moves to constraints.

jfeingold35:
This PR is the implementation of a syntax for command-level constraints. It will close #1507 when it is finally merged.

@W-20534184@

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -278,7 +278,6 @@ export abstract class Command {
     const opts = {
       context: this,
       ...options,
-      constraints: options.constraints,
       flags: aggregateFlags<F, B>(options.flags, options.baseFlags, options.enableJsonFlag),
     }
 
--- a/src/command.ts
+++ b/src/command.ts
@@ -13,6 +13,7 @@ import {
   ArgOutput,
   ArgProps,
   BooleanFlagProps,
+  Constraint,
   Deprecation,
   FlagInput,
   FlagOutput,
@@ -53,6 +54,7 @@ export abstract class Command {
   /** An order-dependent object of arguments for the command */
   public static args: ArgInput = {}
   public static baseFlags: FlagInput
+  public static constraints: Constraint[]
   /**
    * Emit deprecation warning when a command alias is used
    */
@@ -276,6 +278,7 @@ export abstract class Command {
     const opts = {
       context: this,
       ...options,
+      constraints: options.constraints,
       flags: aggregateFlags<F, B>(options.flags, options.baseFlags, options.enableJsonFlag),
     }
 
--- a/src/constraints.ts
+++ b/src/constraints.ts
@@ -16,28 +16,32 @@ export function combinationOf(...flagNames: string[]): FlagGroup {
   }
 }
 
+type ConstraintApplicatorFunction = (flags: FlagOutput) => string
+
 class ConstraintImpl implements Constraint {
   public readonly are: ConstraintImpl = this
   public readonly is: ConstraintImpl = this
   private readonly constrainedFlags: string[]
-  private constraintApplicatorFunction: (flags: FlagOutput) => string | null
-  private constraintType: string
+  private constraintApplicatorFunctionHolder: ConstraintApplicatorFunctionHolder =
+    new ConstraintApplicatorFunctionHolder()
   private topLevelCondition: Condition | undefined
   private underConstructionCondition: Condition | undefined
 
   public constructor(constrainedFlags: string[]) {
     this.constrainedFlags = constrainedFlags
-    this.constraintType = 'default'
-    this.constraintApplicatorFunction = () => null
   }
 
   public get and(): ConstraintImpl {
     if (this.topLevelCondition === undefined) {
-      throw new Error('FILLER: NO CONDITION')
+      throw new Error(
+        `Misconfigured constraint on ${createFlagString(this.constrainedFlags)}: 'and' requires a 'when' or 'unless'.`,
+      )
     }
 
     if (this.underConstructionCondition) {
-      throw new Error('FILLER: CONDITION ALREADY BEING CONSTRUCTED')
+      throw new Error(
+        `Misconfigured constraint on ${createFlagString(this.constrainedFlags)}: 'and' cannot directly follow '${this.underConstructionCondition.getName()}'`,
+      )
     }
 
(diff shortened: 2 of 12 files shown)


src/constraints.ts
```diff
@@ -0,0 +1,741 @@
...
+}
+
+abstract class UnaryOpCondition extends Condition {
+  protected condition: Condition | undefined
+
+  public setCondition(condition: Condition): void {
+    // istanbul ignore if - should be unreachable
+    if (this.condition) {
+      throw new Error('FILLER: DUPLICATE CONDITION')
+    }
```
soridalac, line 614: do this need to throw a message instead?
example:  ` throw new Error(Misconfigured constraint: duplicate condition set on ${this.getName()})   `

src/command.ts
```diff
@@ -276,6 +278,7 @@ export abstract class Command {
     const opts = {
       context: this,
       ...options,
+      constraints: options.constraints,
```
soridalac: add constraints here is redundant of `...options` or is this intentional to override?

in `...options`:
```
export type Input<TFlags extends FlagOutput, BFlags extends FlagOutput, AFlags extends ArgOutput> = {
  flags?: FlagInput<TFlags>
  baseFlags?: FlagInput<BFlags>
  constraints?: Constraint[] | undefined
  enableJsonFlag?: true | false
  args?: ArgInput<AFlags>
  strict?: boolean | undefined
  context?: ParserContext
  '--'?: boolean
}
```

src/constraints.ts
```diff
@@ -0,0 +1,741 @@
...
+ */
+export function flags(...flagNames: string[]): ConstraintImpl {
+  return new ConstraintImpl(flagNames)
+}
+
+/**
+ * Declare a set of flags to be evaluated as one instead of separately.
+ *
+ * @example
+ * flag('foo').is.dependentOn(combinationOf('bar', 'baz))
```
soridalac: mini typo `'baz'`

src/constraints.ts
jfeingold35: In theory this should be unreachable, but yeah, can't hurt to have a specific message here.

src/command.ts
jfeingold35: Oh, good catch. I'll remove the `constraints` declaration here. It serves no purpose.

src/constraints.ts
jfeingold35: Oh good catch!

soridalac (approved):

soridalac:
**QA notes**:
Create a `demo-constraints.ts` flag constraints for parser, ran a demo script covering all constraint types.
Using: `yarn ts-node demo-constraints.ts`

✅ : mutuallyExclusive —` --json` and `--csv` cannot both be used at the same time.
✅ : mutuallyDependent — `--user `and `--password` must be used together or not at all.
✅ : dependentOn — `--output `requires `--format ` to also be present.
✅ : exclusiveWith — `--verbose` cannot be used alongside `--quiet`.
✅ : requiredAny — at least one of `--file`, `--stdin`, or `--url` must be provided.
✅ : requiredAll — both `--host` and `--port` must always be present.
✅ : requiredAtLeastN — at least 2 of `--a`, `--b`, `--c` must be provided.
✅ : conditional (when.thisIsTrue) — `--target` is only required when `--mode` equals deploy.

PR #1613, opened 2026-06-10
fix: causative errors are now also pretty-printed. @W-22797799@

### LINKED ISSUE #1605: Debug-mode error rendering drops the `Error.cause` chain
**Describe the bug**

When `settings.debug` is on, oclif's error renderer returns `error.stack` (in `src/errors/errors/pretty-print.ts` → `prettyPrint`). This shows the outer error's stack only — the standard ES2022 / Node 16.9+ `Error.cause` property is never consulted, so the inner error's message and stack are silently dropped even in the diagnostic-output mode that users opt into specifically when they need more detail.

For comparison, Node's own default uncaught-exception printer recurses the cause chain via `util.inspect` and includes a `[cause]: ...` block.

**To Reproduce**

1. Create a command that wraps an inner error using the standard [`Error(message, { cause })`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) pattern, with `settings.debug` enabled:

   ```ts
   import { Command, settings } from "@oclif/core";
   settings.debug = true;

   export default class Demo extends Command {
     async run() {
       try {
         throw new Error("Inner detail with diagnostic info");
       } catch (err) {
         throw new Error("High-level context", { cause: err });
       }
     }
   }
   ```

2. Run the command.
3. Observe the output.

**Expected behavior**

The cause chain is included, so users get the full diagnostic picture they enabled debug mode to see — for example:

```
Error: High-level context
    at Demo.run (.../demo.mjs:7:10)
    at Demo._run (.../command.js:183:33)
Caused by: Error: Inner detail with diagnostic info
    at Demo.run (.../demo.mjs:5:11)
    at Demo._run (.../command.js:183:33)
```

**Actual behavior**

```
Error: High-level context
    at Demo.run (.../demo.mjs:7:10)
    at Demo._run (.../command.js:183:33)
```

The inner error is absent. Workaround: bypass oclif's renderer entirely and `console.error(err)` before throwing.

**Environment**

- OS & version: macOS 26.5 (Tahoe)
- `@oclif/core` version: 4.11.4 (latest at time of filing)
- Node: 22.x and 24.x (both affected; `Error.cause` available since 16.9)

**Additional context**

Suggested location of fix: in the `settings.debug` branch of `prettyPrint`, walk the `cause` chain and concatenate each level's stack:

```ts
if (settings.debug) {
  const parts = [error.stack];
  let cur: unknown = error.cause;
  while (cur instanceof Error) {
    parts.push(`Caused by: ${cur.stack}`);
    cur = cur.cause;
  }
  return parts.join("\n");
}
```

jfeingold35:
Closes #1605 and @W-22797799@

### DIFF
--- a/src/errors/errors/pretty-print.ts
+++ b/src/errors/errors/pretty-print.ts
@@ -30,28 +30,47 @@ const formatSuggestions = (suggestions?: string[]): string | undefined => {
   return `${label}\n${indent(multiple, 2)}`
 }
 
-export default function prettyPrint(error: Error & PrettyPrintableError & CLIErrorDisplayOptions): string | undefined {
-  if (settings.debug) {
-    return error.stack
+type CombinedErrorType = Error & PrettyPrintableError & CLIErrorDisplayOptions
+
+function isCombinedErrorType(obj: any): obj is CombinedErrorType {
+  return Boolean(obj) && 'name' in obj && 'message' in obj
+}
+
+export default function prettyPrint(error: CombinedErrorType): string | undefined {
+  const prettyPrintedErrors: string[] = []
+  let currentError: unknown = error
+  let isDeep: boolean = false
+  while (isCombinedErrorType(currentError)) {
+    if (settings.debug && currentError.stack) {
+      prettyPrintedErrors.push(`${isDeep ? 'Caused by: ' : ''}${currentError.stack}`)
+    } else {
+      const {bang, code, message, name: errorSuffix, ref, suggestions} = currentError
+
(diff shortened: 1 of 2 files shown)


soridalac (approved):

soridalac:
**QA notes**:
✅ : **Debug Mode Off**: show pretty-printed error messages, user friendly output.

Screenshot 2026-06-15 at 1 00 58 PM

✅ :  **Debug Mode On**: show complete stack traces for all errors in the chain.
Screenshot 2026-06-15 at 1 03 38 PM

PR #1614, opened 2026-06-12
fix: windows shell identification is more reliable @W-22969497@

jfeingold35:
This fixes the @oclif/core tests which were broken on `main` in Windows running on Node v24.16.0.
This is an issue with [Node itself](https://github.com/nodejs/node/issues/63858), wherein the output for `process.title` changed in between node minor versions, which broke some of our code that depends on it having a certain range of values.

 Since we have no timeline on a fix from Node, I workshopped an alternative workaround. This should be more reliable than the previous version at determining what shell is being used on a Windows machine, but it has the downside of doing an `execSync` call, which adds about a half-second of execution time on Windows exclusively.

I'm not sure any alternative workaround exists, so we either need to wait for Node to fix it in their own time, or accept the slight performance hit.

This fixes work item @W-22969497@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -1,7 +1,6 @@
 import * as ejs from 'ejs'
-import {execSync} from 'node:child_process'
-import {arch, userInfo as osUserInfo, release, tmpdir, type} from 'node:os'
-import {join, resolve, sep} from 'node:path'
+import {arch, release, tmpdir, type} from 'node:os'
+import {join, resolve} from 'node:path'
 import {fileURLToPath, URL} from 'node:url'
 
 import Cache from '../cache'
@@ -19,7 +18,7 @@ import {settings} from '../settings'
 import {determinePriority} from '../util/determine-priority'
 import {safeReadJson} from '../util/fs'
 import {toStandardizedId} from '../util/ids'
-import {getHomeDir, getPlatform} from '../util/os'
+import {getHomeDir, getPlatform, getShell} from '../util/os'
 import {compact, isProd} from '../util/util'
 import {ux} from '../ux'
 import {parseTheme} from '../ux/theme'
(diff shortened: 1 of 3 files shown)


src/config/config.ts
```diff
@@ -692,6 +682,20 @@ export class Config implements IConfig {
     }
   }
 
+  private determineWindowsShell(): string {
+    try {
+      const parentProcessName = execSync(
+        `powershell.exe -NoProfile -Command "(Get-CimInstance Win32_Process -Filter 'ProcessID - ${process.ppid}').Name"`,
```
mshanemc: I don't know much about windows to know if this works or not (without booting the personal mac over to windows to find out)

but the approach seems fine (the catch will carry users w/o pwsh access).

--
main objection: this should be a standalone fn in module scope or in another module, not a private method on a class it doesn't use (no `this` references in here)

the class is sprawling enough already

there's an eslint rule to prevent it https://eslint.org/docs/latest/rules/class-methods-use-this

jfeingold35: I did some pretty extensive tests of the little script on my Windows laptop, and it seemed to work fine.
But yeah, I can move the method into a different file. Good call.

mshanemc (approved):

sorccu:
We have a CLI (linked in the PR mentioned above) with an extensive e2e test suite that spawns many instances of our CLI executable. Last week we started seeing our e2e tests consistently time out on Windows, and nothing seemed to make sense, until cpu profile snapshots helped uncover oclif calling execSync, which we then managed to pinpoint to this PR.

Our e2e test suites make hundreds of CLI calls, and each one now has to suffer a delay caused by the synchronous powershell invocation. Now, there may be some OS level throttling involved, but under load, we were seeing just the powershell invocation take several seconds on average, which had a cascading effect and essentially broke the entire test suite. We were able to work around it by manually setting the SHELL environment variable which avoids the powershell call.

I think it would be best to revert this change and study the performance impact more. As it currently stands, the high runtime cost doesn't seem worth the minor improvement in shell detection accuracy.

PR #1615, opened 2026-06-16
fix: include error message in --json output

### LINKED ISSUE #1608: `--json` error rendering drops the error message
**Describe the bug**

When a command with `enableJsonFlag` errors, oclif's `--json` mode serializes the error via `toErrorJson()`, which is `return { error: err }` (in `src/command.ts`). `JSON.stringify` then handles the result — but because `Error.prototype.message`, `.name`, and `.stack` are all non-enumerable own properties, they're skipped. The user-facing error message is lost.

This affects all three idiomatic error patterns: `throw new Error(...)`, `throw new Errors.CLIError(...)`, and `this.error(...)`. Even oclif's first-class `CLIError` doesn't define a `toJSON` to recover the message, so its serialized form has structural fields (`oclif.exit`) but still no message text.

For comparison, Node's own default uncaught-exception printer surfaces the message because it uses `util.inspect`, which has special handling for `Error` instances.

**To Reproduce**

A minimal repro (three commands, one project) is attached as a tarball. Run each command in both modes:

```sh
$ ./bin/run.mjs foo:throw-plain
    Error: This message will be lost under --json

$ ./bin/run.mjs foo:throw-plain --json
{
  "error": {}
}

$ ./bin/run.mjs foo:throw-cli
 ›   Error: This message will ALSO be lost under --json

$ ./bin/run.mjs foo:throw-cli --json
{
  "error": {
    "oclif": {
      "exit": 2
    }
  }
}

$ ./bin/run.mjs foo:use-this-error
 ›   Error: This message will be lost under --json too

$ ./bin/run.mjs foo:use-this-error --json
{
  "error": {
    "oclif": {
      "exit": 2
    }
  }
}
```

The commands themselves are tiny:

```ts
// foo/throw-plain.mjs
import { Command } from "@oclif/core";
export default class ThrowPlain extends Command {
  static enableJsonFlag = true;
  async run() {
    throw new Error("This message will be lost under --json");
  }
}
```

```ts
// foo/throw-cli.mjs
import { Command, Errors } from "@oclif/core";
export default class ThrowCli extends Command {
  static enableJsonFlag = true;
  async run() {
    throw new Errors.CLIError("This message will ALSO be lost under --json");
  }
}
```

```ts
// foo/use-this-error.mjs
import { Command } from "@oclif/core";
export default class UseThisError extends Command {
  static enableJsonFlag = true;
  async run() {
    this.error("This message will be lost under --json too");
  }
}
```

**Expected behavior**

The error message text should appear in the JSON output, e.g.:

```json
{
  "error": {
    "message": "This message will be lost under --json",
    "name": "Error",
    "oclif": { "exit": 2 }
  }
}
```

**Actual behavior**

As shown above — empty object for plain `Error`, structural-only object for `CLIError` / `this.error()`. The message is never present.

**Environment**

- OS & version: macOS 26.5 (Tahoe)
- `@oclif/core` version: 4.11.4 (latest at time of filing)
- Node: 22.x

**Additional context**

Root cause: `Command.toErrorJson(err)` in `src/command.ts` returns `{ error: err }` and relies on `JSON.stringify` for serialization. `JSON.stringify(new Error("foo"))` returns `"{}"` because the Error class's relevant own properties (`message`, `name`, `stack`) are all non-enumerable in V8 / Node.

Suggested location of fix: `Command.toErrorJson` should explicitly include `err.message`, `err.name`, and (optionally) `err.stack` — for instance:

```ts
toErrorJson(err: unknown): unknown {
  if (err instanceof Error) {
    return {
      error: {
        message: err.message,
        name: err.name,
        ...err, // pick up CLIError's `oclif`, `code`, `suggestions` if present
      },
    };
  }
  return { error: err };
}
```

A minimal-project tarball is attached for one-line reproduction (`npm install && ./bin/run.mjs foo:throw-plain --json`).

[oclif-json-error-repro.tar.gz](https://github.com/user-attachments/files/28577196/oclif-json-error-repro.tar.gz)

spokodev:
Closes #1608.

When a command with `enableJsonFlag` errors, `--json` mode serializes the error via `toErrorJson()`, which returned `{error: err}`. `JSON.stringify` (in `logJson`) then drops `Error.prototype.message`/`.name`/`.stack` because they are non-enumerable own properties, so the user-facing message is lost:

```
$ mycmd --json          # throw new Error('...')
{ "error": {} }
$ mycmd --json          # throw new Errors.CLIError('...')
{ "error": { "oclif": { "exit": 2 } } }
```

This affects all three idiomatic patterns: `throw new Error(...)`, `throw new Errors.CLIError(...)`, and `this.error(...)`.

### Fix
In `toErrorJson`, when the error is an `Error`, serialize `message`/`name` explicitly while spreading enumerable own properties, so existing structural fields (e.g. a `CLIError`'s `oclif` metadata) are preserved and the message is surfaced. Non-`Error` values are passed through unchanged.

### Tests
Two `--json error rendering` tests, red before the change and green after — a plain `Error` and a `CLIError` (asserting the message is present and `oclif.exit` is still emitted). Full `test/command` + `test/errors` suites pass.

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -302,6 +302,13 @@ export abstract class Command {
   public abstract run(): Promise<any>
 
   protected toErrorJson(err: unknown): any {
+    if (err instanceof Error) {
+      // `Error.prototype.message`/`.name` are non-enumerable, so they are dropped
+      // by `JSON.stringify` in `logJson`. Surface them explicitly while preserving
+      // any enumerable own properties (e.g. a `CLIError`'s `oclif` metadata).
+      return {error: {...err, message: err.message, name: err.name}}
+    }
+
     return {error: err}
   }
 
--- a/test/command/command.test.ts
+++ b/test/command/command.test.ts
@@ -1,7 +1,7 @@
 import {captureOutput} from '@oclif/test'
 import {expect} from 'chai'
 
-import {Command as Base, Flags} from '../../src'
+import {Command as Base, Errors, Flags} from '../../src'
 
 class Command extends Base {
   static description = 'test command'
@@ -370,6 +370,36 @@ describe('command', () => {
 
       const cmd = new CMD(['--json'], {} as any)
       expect(cmd.jsonEnabled()).to.equal(false)
+    })
+  })
+
+  describe('--json error rendering', () => {
+    it('includes the message of a plain Error', async () => {
+      class CMD extends Command {
+        static enableJsonFlag = true
+
+        async run() {
+          throw new Error('boom under json')
+        }
+      }
+
+      const {stdout} = await captureOutput(async () => CMD.run(['--json']))
+      expect(JSON.parse(stdout).error.message).to.equal('boom under json')
+    })
+
+    it('includes the message of a CLIError alongside its oclif metadata', async () => {
+      class CMD extends Command {
+        static enableJsonFlag = true
+
+        async run() {
+          throw new Errors.CLIError('kaboom under json')
+        }
+      }
+
+      const {stdout} = await captureOutput(async () => CMD.run(['--json']))
+      const json = JSON.parse(stdout)
+      expect(json.error.message).to.equal('kaboom under json')
+      expect(json.error.oclif.exit).to.equal(2)
     })
   })
 })


jfeingold35:
@spokodev , thank you for submitting this. I'll get it reviewed and QA'd, and then assuming nothing needs to change, we can merge it and close the associated Issue.

jfeingold35 (approved):

jfeingold35:
@spokodev , the PR looks good. Assuming the tests all pass, we'll be good to merge it.

PR #1626, opened 2026-07-18
fix: merge execute load options with dir

### LINKED ISSUE #1533: ignoreManifest option in loadOptions not respected by execute()
**Describe the bug**

When using the `execute()` function in development mode with the `ignoreManifest` option set to `true`, the option is not properly passed through to the config loader, resulting in commands still being loaded from the manifest file (`oclif.manifest.json`) instead of from the filesystem. Additionally, when only `ignoreManifest: true` is provided without other options, it causes a `TypeError` due to missing required options that should be inferred automatically.

**To Reproduce**

1. Create a basic oclif CLI with a development entry point (dev.js):

```javascript
#!/usr/bin/env node
import { execute } from '@oclif/core'

await execute({
  development: true,
  dir: import.meta.url,
  loadOptions: {
    ignoreManifest: true,
  }
})
```

2. Modify a command's description in the source file (e.g., add extra characters to make it visually different from the manifest)
3. Run the command: `dev.js your-command --help`
4. Observe the error:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at dirname (node:path:1457:5)
    at up (/node_modules/@oclif/core/lib/util/find-root.js:13:36)
    ...
```

5. To work around the first issue, add `root` option:

```javascript
await execute({
  development: true,
  dir: import.meta.url,
  loadOptions: {
    ignoreManifest: true,
    root: import.meta.dirname
  }
})
```

6. Run the command again: `dev.js your-command --help`
7. Note that the command description matches the **manifest** file, not the source file (the `ignoreManifest` option is not being respected)

**Expected behavior**

1. When `ignoreManifest: true` is provided in `loadOptions`, it should be passed through to the config loader and commands should be loaded from the filesystem
2. Default options (like `root` and `pjson`) should be automatically inferred when `ignoreManifest` is provided, similar to how they're inferred when not using `loadOptions`
3. The command help should display the description from the source file, not from the cached manifest

**Or**, an alternate expected behavior would be when `development: true`, it ignores the manifest so that it is always loading directly from source.

**Screenshots**

Debug output showing the issue:

With `ignoreManifest: true` only:
```
this.options {
  "ignoreManifest": true
}
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```

With both `ignoreManifest: true` and `root`:
```
this.options {
  "ignoreManifest": true,
  "root": "/path/to/bin"
}
ignoreManifest false  ← Still shows false despite being set to true
```

The `ignoreManifest` is logged before this line: [config/plugin.ts](https://github.com/oclif/core/blob/main/src/config/plugin.ts#L268)

 - OS & version: macOS Sequoia 15.2
 - Node.js version: v22.21.1
 - @oclif/core version: 4.8.0

**Additional context**

The issue appears to be in how `execute()` handles the `loadOptions` parameter. It seems like:

1. When `loadOptions` is provided, default options that would normally be inferred are not merged in
2. The `ignoreManifest` option from `loadOptions` is not being passed down to the appropriate Config initialization

This makes it difficult to use development mode with changes to command files not reflected without rebuilding the manifest.

**Workaround**

Currently, there's no clean workaround. Even with explicitly providing `root`, the `ignoreManifest` flag doesn't get passed through, so developers must delete the `oclif.manifest.json` file manually during development to ensure commands load from source files.

liyimil:
## What changed

- merge `execute()`'s `dir` with partial `loadOptions`
- preserve explicitly supplied load options such as `ignoreManifest`
- normalize file URL roots before loading the CLI
- add a regression test covering `dir` plus `{ignoreManifest: true}`

## Why

Passing `loadOptions` currently replaces `dir` entirely, so inferred values such as `root` are lost. Supplying only `ignoreManifest` then reaches the loader without a root and fails before the option can take effect.

Fixes #1533.

## Checks

- `yarn compile --noEmit`
- focused Mocha test on Node 22: `test/execute.test.ts`
- `yarn lint` (passes with existing warnings)
- ESLint and Prettier checks for the changed files

Prepared with Codex assistance; the diff was reviewed and the checks above were run locally.

### DIFF
--- a/src/execute.ts
+++ b/src/execute.ts
@@ -1,9 +1,21 @@
+import {fileURLToPath} from 'node:url'
+
 import {CLIError} from './errors'
 import {handle} from './errors/handle'
 import {flush} from './flush'
-import {LoadOptions} from './interfaces'
+import {LoadOptions, Options} from './interfaces'
 import {run} from './main'
 import {settings} from './settings'
+
+type ExecuteLoadOptions = LoadOptions | (Omit<Options, 'root'> & {root?: string})
+
+function resolveLoadOptions(dir?: string, loadOptions?: ExecuteLoadOptions): LoadOptions {
+  if (!dir || !loadOptions || typeof loadOptions === 'string' || '_base' in loadOptions) {
+    return (loadOptions ?? dir) as LoadOptions
+  }
+
+  return {root: dir.startsWith('file://') ? fileURLToPath(dir) : dir, ...loadOptions}
+}
 
 /**
  * Load and run oclif CLI
@@ -46,7 +58,7 @@ export async function execute(options: {
   args?: string[]
   development?: boolean
   dir?: string
-  loadOptions?: LoadOptions
+  loadOptions?: ExecuteLoadOptions
 }): Promise<unknown> {
   if (!options.dir && !options.loadOptions) {
     throw new CLIError('dir or loadOptions is required.')
@@ -58,7 +70,7 @@ export async function execute(options: {
     settings.debug = true
   }
 
-  return run(options.args ?? process.argv.slice(2), options.loadOptions ?? options.dir)
+  return run(options.args ?? process.argv.slice(2), resolveLoadOptions(options.dir, options.loadOptions))
     .then(async (result) => {
       flush()
       return result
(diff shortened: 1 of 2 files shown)


jfeingold35:
Closed and re-opened to re-evaluate the CLA.
I will review this, conduct any necessary QA, and merge if it's ready to go. Thank you for the contribution, @liyimil.

jfeingold35:
@liyimil , this looks like it resolves the issue where you needed to have `root`, but it still seems like the description is being pulled from the manifest instead of the file.

PR #1635, opened 2026-08-06
feat: upping to eslint 10 (W-23473845)

jfeingold35:
Part of @W-23473845@.

Bumping `eslint` to v10 and `eslint-config-oclif` to v7.x.

Running `yarn lint --fix` introduced compile errors (which were easily fixed) and test failures (which were arcane nonsense that stymied me for hours). So I just decided to flatly suppress every violation.

Suppression was done through a suppression.json file that grandfathers in all the existing violations.


src/parser/parse.ts
```diff
@@ -53,7 +53,7 @@ declare global {
   var oclif: {stdinCache?: string}
 }
 
-export const readStdin = async (): Promise<null | string> => {
+export const readStdin = async (): Promise<undefined | string> => {
```
jfeingold35: I had to manually change some `return null`s to `return undefined`. I determined that the `export` here is _solely_ for testability, and all the internal usages are just `if (!blah)`, so there's nothing to worry about with behavioral changes.

src/interfaces/plugin.ts
```diff
@@ -36,13 +36,9 @@ export interface Options extends PluginOptions {
...
   userPlugins?: boolean | undefined
   version?: string | undefined
-}
+} & PluginOptions
 
-export interface Plugin {
-  /**
-   * ../config version
-   */
-  _base: string
```
jfeingold35, line 45: `config/plugin.ts` changed a property that was implicitly private (`_base`) to one that is _explicitly_ private (`#base`), but that introduced a compile error due to the definition here.
It is Claude's (and also my) opinion that removing this property is safe to do, because it was not referenced anywhere except for internally, and it's an implicitly private property anyway.

src/config/config.ts
```diff
@@ -514,7 +523,7 @@ export class Config implements IConfig {
         },
       }
 
-      const hooks = p.hooks[event] || []
+      const hooks = p.hooks[event as string] || []
```
jfeingold35: Had to manually add this to resolve another compile error.

eslint.config.mjs
```diff
@@ -6,6 +6,7 @@ import {fileURLToPath} from 'node:url'
 
 const gitignorePath = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '.gitignore')
 
+// eslint disable-next-line import-x/no-anonymous-default-export
```
jfeingold35, line 9: I don't really know why I manually suppressed this here instead of using the suppression file. It felt weird to implicitly suppress something in the eslint config itself. I want to call it "rude" for some reason?

WillieRuemmele (dismissed):

WillieRuemmele (approved):

PR #1640, opened 2026-08-18
fix: ignore defaulted flags in dependency validation

### LINKED ISSUE #1639: dependsOn / relationships type 'all' is satisfied by a flag's own default, unlike exclusive and combinable
**Version:** `@oclif/core` 4.8.0

## Summary

`dependsOn` (and `relationships: [{type: 'all', ...}]`, which routes to the same validator) treats a flag as "provided" whenever it has a defined value after parsing. That includes values that came from the flag's own `default`. As a result, a dependency on any flag that declares a `default` can never fail — the check silently always passes.

This is inconsistent with the exclusivity validators in the same file, which explicitly skip defaulted flags.

## The code

From `lib/parser/validate.js` in 4.8.0, `validateExclusive` (and identically `validateCombinable`) skips flags that were set from a default:

```js
async function validateExclusive(name, flags) {
    const base = { name, validationFn: 'validateExclusive' };
    const resolved = await resolveFlags(flags);
    const keys = getPresentFlags(resolved);
    for (const flag of keys) {
        // do not enforce exclusivity for flags that were defaulted
        if (parse.output.metadata.flags && parse.output.metadata.flags[flag]?.setFromDefault)
            continue;
        if (parse.output.metadata.flags && parse.output.metadata.flags[name]?.setFromDefault)
            continue;
        ...
```

`validateDependsOn` has no such check — it only tests for `!== undefined`:

```js
async function validateDependsOn(name, flags) {
    const base = { name, validationFn: 'validateDependsOn' };
    const resolved = await resolveFlags(flags);
    const foundAll = Object.values(resolved).every((val) => val !== undefined);
    if (!foundAll) {
        ...
```

`resolveFlags` reads values straight out of `parse.output.flags`, which already has defaults applied, so a defaulted flag is always `!== undefined`.

And `relationships` of type `'all'` dispatches to that same function:

```js
function validateRelationships(name, flag) {
    return (flag.relationships ?? []).map((relationship) => {
        switch (relationship.type) {
            case 'all': {
                return validateDependsOn(name, relationship.flags);
            }
            case 'none': {
                return validateExclusive(name, relationship.flags);
            }
            case 'only': {
                return validateCombinable(name, relationship.flags);
            }
            ...
```

So the `setFromDefault` distinction — "set from a default" is not the same as "the user provided this" — is applied to three of the four relationship types, but not to `all`/`dependsOn`.

## Reproduction

```ts
static flags = {
  mode: Flags.string({ options: ['a', 'b'], default: 'a' }),
  extra: Flags.string({ dependsOn: ['mode'] }),
}
```

Running:

```
mycmd --extra=x
```

**Expected:** validation fails — `--mode` was never supplied, and the author's intent was "if you pass `--extra` you must also pass `--mode`".

**Actual:** validation passes, because `mode` defaulted to `'a'` and is therefore `!== undefined`.

The same happens with the `relationships` spelling:

```ts
extra: Flags.string({ relationships: [{type: 'all', flags: ['mode']}] }),
```

## Impact

The failure is silent. There is no warning at definition time and no error at runtime — the constraint simply never fires, so a CLI author believes they have validation they do not have. Any `dependsOn` target that carries a `default` is unenforceable.

We hit this in a production CLI and only discovered it by reading oclif's source after the constraint failed to reject a combination we expected it to reject.

## Possible resolutions

Either would work for us:

1. Apply the same `setFromDefault` skip in `validateDependsOn`, for consistency with `validateExclusive` / `validateCombinable`.
2. Or, if defaults are intentionally treated as provided for dependencies, document that on `dependsOn` / `relationships` at https://oclif.io/docs/flags. The current docs say only "Make this flag dependent on all of these flags", which reads as a statement about what the user passed.

Happy to send a PR for whichever direction you prefer.

liujiayang2026:
## Summary

- treat dependency flags populated only by defaults as not provided
- preserve dependency validation when an explicit value matches the flag default
- add regression coverage for `dependsOn` and `relationships` with type `all`

## Testing

- `yarn mocha --forbid-only test/parser/parse.test.ts test/parser/validate.test.ts`
- `yarn test`

Closes #1639

### DIFF
--- a/src/parser/validate.ts
+++ b/src/parser/validate.ts
@@ -274,7 +274,9 @@ export async function validate(parse: {input: ParserInput; output: ParserOutput}
     const base = {name, validationFn: 'validateDependsOn'}
     const resolved = await resolveFlags(flags)
 
-    const foundAll = Object.values(resolved).every((val) => val !== undefined)
+    const foundAll = Object.entries(resolved).every(
+      ([flag, value]) => value !== undefined && !parse.output.metadata.flags?.[flag]?.setFromDefault,
+    )
     if (!foundAll) {
       const formattedFlags = Object.keys(resolved)
         .map((f) => `--${f}`)
--- a/test/parser/parse.test.ts
+++ b/test/parser/parse.test.ts
@@ -1575,6 +1575,52 @@ See more help with --help`)
 
       expect(message).to.include('All of the following must be provided when using --foo: --bar')
     })
+
+    it('fails when a dependency is set from its default', async () => {
+      let message = ''
+      try {
+        await parse(['--foo', 'a'], {
+          flags: {
+            foo: Flags.string({dependsOn: ['bar']}),
+            bar: Flags.string({default: 'b'}),
+          },
+        })
+      } catch (error: any) {
+        message = error.message
+      }
+
+      expect(message).to.include('All of the following must be provided when using --foo: --bar')
+    })
+
+    it('succeeds when a dependency explicitly matches its default', async () => {
+      const out = await parse(['--foo', 'a', '--bar', 'b'], {
+        flags: {
+          foo: Flags.string({dependsOn: ['bar']}),
+          bar: Flags.string({default: 'b'}),
+        },
+      })
+
+      expect(out.flags.foo).to.equal('a')
+      expect(out.flags.bar).to.equal('b')
+    })
+  })
+
+  describe("relationships: 'all'", () => {
+    it('fails when a dependency is set from its default', async () => {
+      let message = ''
+      try {
+        await parse(['--foo', 'a'], {
+          flags: {
+            foo: Flags.string({relationships: [{flags: ['bar'], type: 'all'}]}),
+            bar: Flags.string({default: 'b'}),
+          },
+        })
+      } catch (error: any) {
+        message = error.message
+      }
+
+      expect(message).to.include('All of the following must be provided when using --foo: --bar')
+    })
   })
 
   describe('exclusive', () => {


liujiayang2026:
I have signed the Salesforce CLA, and the official CLA status page now reports that all external contributors have signed. The `salesforce-cla` commit status and `cla:missing` label still appear to be stale; could a maintainer please refresh or re-run the CLA check for the current head SHA `458faf5280cd5bbc986aaa658fc14144ccc4ec43`?

WillieRuemmele (dismissed):
Reviewed the fix — it's correct and consistent with how `validateExclusive` and `validateCombinable` already handle `setFromDefault` flags. All tests pass (including the 3 new ones), full suite clean.

One minor observation for a follow-up: this checks if the dependency *target* was `setFromDefault`, but not if the flag *with* the `dependsOn` constraint was itself `setFromDefault` (which `validateExclusive` does check at both ends). That's a separate edge case — not a blocker here.

@W-24192296@

WillieRuemmele:
Thanks for the detailed investigation and the clean PR, @liujiayang2026.

After reviewing this, we've concluded that the current `dependsOn` behavior — treating flags with defaults as "present" — is consistent and likely intentional rather than a bug:

1. **Consistent across APIs**: The newer constraints API (`flag('foo').is.dependentOn('bar')`) has the same semantics — `filterFlagsPresentInInput` checks `!== undefined` without considering `setFromDefault`. Both APIs treat "has a value" as "is present."

2. **Different semantics from `exclusive`**: `exclusive`/`combinable` skip defaulted flags because they're about user intent ("don't pass conflicting flags"). `dependsOn` is about functional dependency ("this flag needs that flag to have a value"), which defaults satisfy.

3. **Breaking change**: Tightening this validation would reject previously-accepted input, which is a semver breaking change regardless of intent.

We'll consider this for the **next major version** where we can revisit the semantics holistically across both `dependsOn` and the constraints API.

**Workaround for now**: If you need "user must explicitly pass this flag" semantics, you can use the [constraints API](https://oclif.io/docs/flags#custom-constraint-api) with a `.when.thisIsTrue()` condition, or move the default out of the flag definition and apply it in your `run()` method so that `dependsOn` sees `undefined` when the user doesn't pass the flag.

